PR #27, opened 2021-06-25
Base learning rate decay on training loss

peastman:
This changes it to base learning rate decay on training loss rather than validation loss.  That gives a much cleaner signal for whether it is still learning.  The practical effect is that you can use a smaller value for `lr_patience`, which leads to faster training.

In general, training loss tells you whether it is learning, and the difference between training loss and validation loss tells you whether it is overfitting.  If the training loss stops decreasing, that means you need to reduce the learning rate.  If the training loss is still decreasing but the validation loss stops going down, that means it is overfitting and you should stop.  Reducing the learning rate won't help.

### DIFF
--- a/torchmdnet/module.py
+++ b/torchmdnet/module.py
@@ -32,7 +32,7 @@ class LNNP(LightningModule):
                                       patience=self.hparams.lr_patience,
                                       min_lr=self.hparams.lr_min)
         lr_scheduler = {'scheduler': scheduler,
-                        'monitor': 'val_loss',
+                        'monitor': 'train_loss',
                         'interval': 'epoch',
                         'frequency': 1}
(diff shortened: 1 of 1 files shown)


peastman:
I want to make another push for this change.  Can we at least make it a supported option?  It makes a huge difference to training speed.

I consistently find that the fastest way to train models is to start with a high learning rate, then reduce it aggressively when training stalls (`lr_patience` set to 1 or 2 and `lr_factor` around 0.5).  But that only works when basing it on training loss.  Validation loss is much too noisy, so you have to wait much longer before reducing the learning rate.  That leads to much slower training

giadefa:
I am fine with that, if we can make it an option.

peastman:
This is superseded by #89.

PR #51, opened 2022-01-10
Fix the initialization of CFConv

raimis:


raimis:
@PhilippThoelke could you review?

PhilippThoelke (approved):

PR #52, opened 2022-01-13
Enable more CI triggers

raimis:
- [x] Test merges to `main`
- [x] Test each PR
- [x] Test `main` weakly (every Sunday at 0:00)


raimis:
@PhilippThoelke @stefdoerr could you check?

PR #50, opened 2022-01-10
 Accelerate the limited TorchMD_GN with NNPOps

raimis:
- [x] Integrate `NNPOps`
- [x] Tests
- [x] Benchmarks

~~Depend on https://github.com/openmm/NNPOps/pull/40/~~

PR #55, opened 2022-02-17
Update the installation instructions

raimis:
- [x] Update the installation instructions

This is in preparation of #50, where everything has to be compatible with `conda-forge`, otherwise bad things are happening.


raimis:
@stefdoerr @PhilippThoelke could you check?

PhilippThoelke:
Does installing pytorch just from the environment yaml file work? The official installation instructions are more complex than that: https://pytorch.org/get-started/locally/
Before we referenced this in the readme.

raimis:
> Does installing pytorch just from the environment yaml file work?

Yes, I use the same environment file for the CI.

> The official installation instructions are more complex than that: https://pytorch.org/get-started/locally/

The official packages are not compatible with `conda-forge`.

PhilippThoelke (approved):

PR #62, opened 2022-03-09
Convert the unknown `load_model` parameters to warnings

raimis:
This to enable to load the old check point files:

```python
model = load_model('old.ckpt', aggr='add')
```


raimis:
@PhilippThoelke @stefdoerr can you review?

PhilippThoelke (approved):

PR #63, opened 2022-03-17
Add support for the Ace dataset and ability to pass the charge and spin state of molecules

raimis:
- [x] Implement a dateset
- [x] Integrate
  - [x] Dataset
  - [x] Charges and spins

### DIFF
--- a/scripts/train.py
+++ b/scripts/train.py
@@ -64,6 +64,8 @@ def get_args():
     parser.add_argument('--prior-model', type=str, default=None, choices=priors.__all__, help='Which prior model to use')
 
     # architectural args
+    parser.add_argument('--charge', type=bool, default=False, help='Model needs a total charge')
+    parser.add_argument('--spin', type=bool, default=False, help='Model needs a spin state')
     parser.add_argument('--embedding-dimension', type=int, default=256, help='Embedding dimension')
     parser.add_argument('--num-layers', type=int, default=6, help='Number of interaction layers in the model')
     parser.add_argument('--num-rbf', type=int, default=64, help='Number of radial basis functions in model')
--- a/tests/test_equivariance.py
+++ b/tests/test_equivariance.py
@@ -18,8 +18,8 @@ def test_scalar_invariance():
     pos = torch.randn(100, 3)
     batch = torch.arange(50, dtype=torch.long).repeat_interleave(2)
 
-    y = model(z, pos, batch=batch)[0]
-    y_rot = model(z, pos @ rotate, batch=batch)[0]
+    y = model(z, pos, batch)[0]
+    y_rot = model(z, pos @ rotate, batch)[0]
     torch.testing.assert_allclose(y, y_rot)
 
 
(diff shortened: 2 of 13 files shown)


jule-c (approved):

raimis:
I have tried to train and seems to work:
```
activation: silu
aggr: add
atom_filter: -1
attn_activation: silu
batch_size: 128
coord_files: null
cutoff_lower: 0.0
cutoff_upper: 0.5 # nm
dataset: Ace
dataset_arg: null
dataset_root: ../orbnet/denali/computation/orbnet_denali_compute_*.h5
derivative: true
distance_influence: both
distributed_backend: ddp
early_stopping_patience: 500
ema_alpha_dy: 1.0
ema_alpha_y: 1.0
embed_files: null
embedding_dimension: 128
energy_files: null
energy_weight: 0.2
force_files: null
force_weight: 0.8
inference_batch_size: 128
load_model: null
log_dir: logs/
lr: 1.0e-3
lr_factor: 0.5
lr_min: 1.0e-07
lr_patience: 5
lr_warmup_steps: 10000
max_num_neighbors: 32
max_z: 100
model: graph-network
neighbor_embedding: true
ngpus: -1
num_epochs: 3000
num_heads: 8
num_layers: 6
num_nodes: 1
num_rbf: 32
num_workers: 6
output_model: Scalar
precision: 32
prior_model: null
rbf_type: expnorm
redirect: false
reduce_op: add
save_interval: 10
splits: null
standardize: false
test_interval: 10
test_size: null
train_size: 0.8
trainable_rbf: true
val_size: 0.05
weight_decay: 0.0
```

raimis:
@PhilippThoelke and @jule-c could you review? Especially the part of passing charges and spin state to the models.

raimis:
@PhilippThoelke that was my initial idea to use `**kwargs`, but TorchScript doesn't like them. I cannot pass the test neither with `**kwargs`, nor name arguments:
https://github.com/torchmd/torchmd-net/blob/23f479063943b985ab05448b774d93f8c3d7fd2a/tests/test_model.py#L32-L38

Any ideas how to circumvent that?

PhilippThoelke (changes requested):
Ah right, forgot about TorchScript. I would still suggest to maintain the previous order of parameters, i.e. `z, pos, batch` and only include charge and spin as keyword arguments after that. Currently the change is not backwards compatible, producing a hard to debug error. Wherever the batch is passed as a positional argument (also outside this repository), the model won't throw an error but instead run a prediction without batching. Without much looking for example I already found this which would break under the current changes and does not seem to be covered by a test. https://github.com/torchmd/torchmd-net/blob/23f479063943b985ab05448b774d93f8c3d7fd2a/torchmdnet/calculators.py#L18

I can't think of a TorchScript compatible solution that is more general than your current suggestion, however, I don't think this is ideal. Especially if there will be even more attributes in the future, this `if` will increase exponentially in the number of cases. https://github.com/torchmd/torchmd-net/blob/27395e446cd8b735c66897bc93b52b4c09b8c73c/torchmdnet/module.py#L75-L82

raimis:
@PhilippThoelke the issue with the `AtomFilter` is fixed. Could you review again?

PhilippThoelke (approved):
Looks good!

PhilippThoelke:
Just realized that the additional arguments break the AtomFilter at another place, I'll add a quick fix and tests on main.

PR #67, opened 2022-03-29
fix bugs for the newest version of pytorch_lightning

lsnty5190:
sync for pytorch-lightning 1.5.10:

1. trainer.hparams -> save_hyperparameters
2. trainer.reload_dataloaders_every_epoch -> trainer.reload_dataloaders_every_n_epochs
3. trainer.running_sanity_check -> trainer.sanity_checking
4. periods -> every_n_epochs

### DIFF
--- a/torchmdnet/data.py
+++ b/torchmdnet/data.py
@@ -13,7 +13,8 @@ from torch_scatter import scatter
 class DataModule(LightningDataModule):
     def __init__(self, hparams, dataset=None):
         super(DataModule, self).__init__()
-        self.hparams = hparams.__dict__ if hasattr(hparams, "__dict__") else hparams
+        # self.hparams = hparams.__dict__ if hasattr(hparams, "__dict__") else hparams
+        self.save_hyperparameters(hparams.__dict__ if hasattr(hparams, "__dict__") else hparams)
         self._mean, self._std = None, None
         self._saved_dataloaders = dict()
         self.dataset = dataset
@@ -84,7 +85,7 @@ class DataModule(LightningDataModule):
 
     def _get_dataloader(self, dataset, stage, store_dataloader=True):
         store_dataloader = (
-            store_dataloader and not self.trainer.reload_dataloaders_every_epoch
+            store_dataloader and not self.trainer.reload_dataloaders_every_n_epochs
         )
         if stage in self._saved_dataloaders and store_dataloader:
             # storing the dataloaders like this breaks calls to trainer.reload_train_val_dataloaders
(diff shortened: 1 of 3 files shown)


torchmdnet/data.py
```diff
@@ -84,7 +85,7 @@ def std(self):
 
     def _get_dataloader(self, dataset, stage, store_dataloader=True):
         store_dataloader = (
-            store_dataloader and not self.trainer.reload_dataloaders_every_epoch
+            store_dataloader and not self.trainer.reload_dataloaders_every_n_epochs
```
peastman, line 88: I would make this

```python
store_dataloader and self.trainer.reload_dataloaders_every_n_epochs == 0
```

to clarify that reload_dataloaders_every_n_epochs is an int, not a bool.

raimis:
@PhilippThoelke do you see any other issues?

raimis:
@lsnty5190 would you mind to update the version (https://github.com/torchmd/torchmd-net/blob/main/environment.yml), so the CI can pass?

PhilippThoelke (changes requested):
@lsnty5190 why are you moving `train.py` outside of the scripts directory?

The CI will probably show a long list of errors when you bump pytorch-lightning to the most recent version in the environment.yml file. I tried running train.py with the new pl version and there are lot's of problems:
- this parameter has changed https://github.com/torchmd/torchmd-net/blob/b9785d203d0f1e30798db44c1abfeeb7f5dc2eac/scripts/train.py#L161
- the trainer arguments `accelerator` and `devices` changed in their behaviour, this probably needs some work
- this plugin is deprecated https://github.com/torchmd/torchmd-net/blob/b9785d203d0f1e30798db44c1abfeeb7f5dc2eac/scripts/train.py#L149
- it seems like pl changed their way of distributing the model to different processes, I get an unpickle error due to the way pytorch-geometric dynamically generates code for jittable message passing modules

I have started porting torchmd-net to a more recent pl version in the past but didn't get very far. I remember there also were some problems with the current setup of validation and testing dataloaders.

It would be great to have torchchmd-net running with recent pl versions, thanks for looking into it!

PhilippThoelke:
The branch where I started moving to a more recent pytorch-lightning version is still there (I did this 8 months ago though, stuff has changed in torchmd-net and pl since then but maybe it still helps): https://github.com/torchmd/torchmd-net/tree/new_pl

Have a look at this for a list of changes that I made: https://github.com/torchmd/torchmd-net/compare/new_pl

PhilippThoelke:
I made a new attempt at upgrading the code base to pytorch-lightning 1.6.3 (most recent version currently) and some of the problems we had earlier seem to be gone now. Closing this now as it is superseded by #97.

PR #68, opened 2022-03-29
Implement neighbors calculation

claudi:
See #61


claudi:
The performance results show that, the executions of both inference and training take less time. Most notably, from 90% of execution time is spent on the actual inference calculations of the model, up from only about 55% (70% for small molecules). This is more pronounced in big molecules/loads, where the percentage of time spent performing the actual calculations goes up to 98%.

This means execution time is now dedicated almost purely to model evaluation (which hasn't changed), rather than auxiliary computations.

It makes sense that the effect is less pronounced in small molecules (although still satisfactory), as the CPU implementation on those cases is fast enough, and the GPU implementation loses non-neglibigle time to communication. As mentioned, it is still faster.

This is as measured by profiling `TorchMD_GN.forward_call` on metro16.

The results are equivalent, up to a tolerance (10e-5) in the distances, from the original implementation, which was the desired behaviour.

raimis:
Could you post a table with the raw numbers of your benchmarks?

claudi:
@raimis have you had the chance to look at this?

raimis:
Yes, the speed up of DHFR and FC9 looks very good.

PR #70, opened 2022-04-11
adds ANI1x as dataset option

sebastianmdick:
I took the self_energies from the ANI1 datasets. Given that ANI1x uses a different XC functional, the atomic 'self energies' should in principle be adjusted to reflect that.

### DIFF
--- a/torchmdnet/datasets/__init__.py
+++ b/torchmdnet/datasets/__init__.py
@@ -1,4 +1,3 @@
-from torchmdnet.datasets.ani1x import ANI1X
 from .qm9 import QM9
 from .md17 import MD17
 from .ani1 import ANI1
--- a/torchmdnet/datasets/__init__.py
+++ b/torchmdnet/datasets/__init__.py
@@ -1,8 +1,10 @@
+from torchmdnet.datasets.ani1x import ANI1X
 from .qm9 import QM9
 from .md17 import MD17
 from .ani1 import ANI1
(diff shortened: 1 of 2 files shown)


torchmdnet/datasets/ani1x.py
```diff
@@ -0,0 +1,97 @@
...
+au_to_eV = 27.211386246
+
+
+class ANI1X(InMemoryDataset):
+
+    raw_url = "https://figshare.com/ndownloader/files/18112775"
+
+    element_numbers = {"H": 1, "C": 6, "N": 7, "O": 8}
+
+    self_energies = {
```
raimis, line 19: Could you add a comment the atomic energies are from the ANI-1 dataset.

torchmdnet/datasets/__init__.py
```diff
@@ -1,8 +1,10 @@
+from torchmdnet.datasets.ani1x import ANI1X
```
PhilippThoelke, line 1: Why import it twice?

PhilippThoelke (commented):
Looks good but I'm confused by the double import in `__init__.py`.

torchmdnet/datasets/__init__.py
sebastianmdick: I think that was added by vscode. I must have missed it.

raimis (approved):

PR #73, opened 2022-04-19
Memory-efficient ANI-1 dataset with MemMap

raimis:
Memory-efficient ANI-1 dataset

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -19,6 +19,7 @@ dependencies:
   - raimis::pytorch-geometric==2.0.2
   - pytorch-lightning==1.3.8
   - setuptools==59.5.0 # Pin due to https://github.com/pytorch/pytorch/pull/69904
+  - torchmetrics <0.8
   - tqdm
   # Dev tools
   - flake8
--- a/torchmdnet/datasets/ani1.py
+++ b/torchmdnet/datasets/ani1.py
@@ -50,9 +50,9 @@ class ANI1(Dataset):
             molecules = list(h5py.File(path).values())[0].values()
 
             for mol in tqdm(molecules, desc='Molecules', leave=False):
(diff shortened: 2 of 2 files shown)


raimis:
**Benchmark**

Load batches of 1024 samples with 4 workers:
```python
from tqdm import tqdm
from torchmdnet.datasets import ANI1
from torch_geometric.loader import DataLoader

ds = ANI1('.')
dl = DataLoader(dataset=ds, batch_size=1024, num_workers=4)

for b in tqdm(dl):
    pass
```

The old implementation:
  - Speed: 25 batch/s
  - Memory: ~16 GB per worker
  
This implementation:
  - Speed: 45 batch/s
  - Memory: ~1 GB per worker

raimis:
@PhilippThoelke can you review?

environment.yml
```diff
@@ -19,6 +19,7 @@ dependencies:
   - raimis::pytorch-geometric==2.0.2
   - pytorch-lightning==1.3.8
   - setuptools==59.5.0 # Pin due to https://github.com/pytorch/pytorch/pull/69904
+  - torchmetrics <0.8
```
PhilippThoelke, line 22: Why are you adding torchmetrics here?

raimis: It is a dependency of PyTorch Lightning, but the latest versions (0.8) is no longer compatible with the old PyL, which we are using.

giadefa:
We should be careful not to save the mem maps on a network drive (e.g.
inside a cluster). Maybe we should allow the user to specify the path?

raimis:
```python
ds = ANI1('.')
```
One can specify a path where the maps are stored.

giadefa:
ok, maybe have something that say that it should be a fast drive.

PhilippThoelke (approved):

PR #69, opened 2022-04-01
POC: message passing

raimis:
This is a proof-of-concept. DO NOT MERGER!

- [x] Python wrapper
- [x] CPU
  - [x] Forward pass
- [x]  CUDA
  - [x] Forward pass
  - [x] Backward pass
- [ ] Tests
- [ ] Documentation

PR #74, opened 2022-04-20
POC: the graph network optimization with the custom kernels

raimis:
- [x] Integrate the neighbor kernel
- [x] Integrate the message kernel
- [x] Benchmark 

Depend on #61 and #69


raimis:
This is obsolted

PR #78, opened 2022-05-04
Dynamic max num neighbors

PhilippThoelke:
This sets the `max_num_neighbors` argument to the `radius_graph` function to a value that makes sure to never miss any neighbors. The value would be the number of atoms of the largest molecule in the batch.


torchmdnet/models/utils.py
```diff
@@ -227,6 +225,9 @@ def forward(self, pos, batch):
             edge_weight = torch.norm(edge_vec, dim=-1)
 
         lower_mask = edge_weight >= self.cutoff_lower
+        if self.loop and mask is not None:
+            # keep self loops even though they might be below the lower cutoff
+            lower_mask = lower_mask | ~mask
```
stefdoerr, line 230: Is this also part of the PR?

tests/test_model_utils.py
```diff
@@ -0,0 +1,59 @@
...
+from pytest import mark, raises
+import torch
+from torchmdnet.models.utils import Distance
+
+
+@mark.parametrize("cutoff_lower", [0, 2])
+@mark.parametrize("cutoff_upper", [5, 10])
+@mark.parametrize("return_vecs", [False, True])
+@mark.parametrize("loop", [False, True])
+def test_make_splits_ratios(cutoff_lower, cutoff_upper, return_vecs, loop):
```
raimis, line 11: The name of the test is strange.

PhilippThoelke: Good catch, copy paste mistake. I've changed it now.

torchmdnet/models/utils.py
PhilippThoelke: It's something else, I'll move it into a separate PR.

giadefa:
we are not merging this. We will keep it as it is but error out if the max
number of neighbors is passed

Issue #77, opened 2022-05-02
NaN when fitting with derivative


peastman:
I'm trying to fit an equivariant transformer model.  If I specify `derivative: true` in the configuration file to use derivatives in fitting, then after only a few training steps the model output becomes nan.  This happens even if I also specify `force_weight: 0.0`.  The derivatives shouldn't affect the loss at all in that case, yet it still causes fitting to fail.  The obvious explanation would be if I had a nan in the training data somewhere, since that would cause the loss to also be nan even after multiplying by 0.  But I verified that's not the case.  Immediately after it computes the loss

https://github.com/torchmd/torchmd-net/blob/b9785d203d0f1e30798db44c1abfeeb7f5dc2eac/torchmdnet/module.py#L89

I added

```python
print(loss_dy, torch.all(torch.isfinite(deriv)), torch.all(torch.isfinite(batch.dy)))
```

Here's the relevant output from the log.

```
Epoch 0:   1%|          | 31/5483 [00:06<19:36,  4.63it/s, loss=1.28e+07, v_num=_]tensor(11670.3730, device='cuda:0', grad_fn=<MseLossBackward0>) tensor(True, device='cuda:0') tensor(True, device='cuda:0')
Epoch 0:   1%|          | 32/5483 [00:06<19:32,  4.65it/s, loss=1.25e+07, v_num=_]tensor(273794.6562, device='cuda:0', grad_fn=<MseLossBackward0>) tensor(True, device='cuda:0') tensor(True, device='cuda:0')
Epoch 0:   1%|          | 33/5483 [00:07<19:28,  4.67it/s, loss=1.25e+07, v_num=_]tensor(nan, device='cuda:0', grad_fn=<MseLossBackward0>) tensor(False, device='cuda:0') tensor(True, device='cuda:0')
Epoch 0:   1%|          | 34/5483 [00:07<19:25,  4.68it/s, loss=nan, v_num=_]     tensor(nan, device='cuda:0', grad_fn=<MseLossBackward0>) tensor(False, device='cuda:0') tensor(True, device='cuda:0')
```

`batch.dy` never contains a non-finite value.

Any idea what could be causing this?

PhilippThoelke:
[...]

PR #79, opened 2022-05-04
Skip gradients of atoms outside cutoff

PhilippThoelke:
This addresses #77 by detaching gradients from atoms that are outside the cutoff radius of any other atom and are therefore unable to interact inside the model. Previously this caused gradients of the force loss to become NaN, now these atoms are ignored and a warning is raised.


peastman:
When I run it, it fails with this error:

```
Traceback (most recent call last):
  File "/home/peastman/workspace/torchmd-net/scripts/train.py", line 11, in <module>
    from torchmdnet import datasets, priors, models
  File "/home/peastman/workspace/torchmd-net/torchmdnet/datasets/__init__.py", line 2, in <module>
    from .ani import ANI1, ANI1CCX, ANI1X
  File "/home/peastman/workspace/torchmd-net/torchmdnet/datasets/ani.py", line 165
    if data := self.filter_and_pre_transform(data):
             ^
SyntaxError: invalid syntax
```

I'm using Python 3.7.  It looks like the code now requires 3.8 or later?  I tried to upgrade Python, but ran into lots conda conflicts.  It looks like the only way I can do that is to recreate my whole environment from scratch.  Is there a good reason from dropping 3.7 support?  The code in question is trivial.  Just change

```python
if data := self.filter_and_pre_transform(data):
    yield data
```

to

```python
data = self.filter_and_pre_transform(data)
if data:
    yield data
```

peastman:
I reinstalled everything with Python 3.9.  I can confirm this fixes the problem.  Thanks!

PhilippThoelke:
I don't really see a reason to drop python 3.7 support. @raimis was there any specific reason to not use older python syntax here? https://github.com/torchmd/torchmd-net/blob/31d5cb80f98db0c816aa6d0f2e97c515560c183a/torchmdnet/datasets/ani.py#L165

I will merge this pull request then.

PR #81, opened 2022-05-04
Throw error if max_num_neighbors is too small

PhilippThoelke:
This adds an assertion to the neighbor search, which makes sure that no atoms are being missed due to a too small `max_num_neighbors` parameter.

### DIFF
--- a/torchmdnet/models/utils.py
+++ b/torchmdnet/models/utils.py
@@ -214,8 +214,17 @@ class Distance(nn.Module):
             r=self.cutoff_upper,
             batch=batch,
             loop=self.loop,
-            max_num_neighbors=self.max_num_neighbors,
+            max_num_neighbors=self.max_num_neighbors + 1,
         )
+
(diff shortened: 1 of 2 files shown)


torchmdnet/models/utils.py
```diff
@@ -214,8 +214,17 @@ def forward(self, pos, batch):
...
-            max_num_neighbors=self.max_num_neighbors,
+            max_num_neighbors=self.max_num_neighbors + 1,
         )
+
+        # make sure we didn't miss any neighbors due to max_num_neighbors
+        assert not (
+            torch.unique(edge_index[0], return_counts=True)[1] > self.max_num_neighbors
+        ).any(), (
+            "The neighbor search missed some atoms due to max_num_neighbors being too low. "
+            "Please increase this parameter to fit your largest molecule."
```
giadefa, line 225: Please increase this parameter to include the maximum atoms within the cutoff.

raimis (approved):

giadefa (approved):

Issue #82, opened 2022-05-05
Unable to fit model


peastman:
I've been trying to train a model on an early subset of the SPICE dataset.  All my efforts so far have been unsuccessful.  I must be doing something wrong, but I really don't know what.  I'm hoping someone else can spot the problem.  My configuration file is given below.  Here's the [HDF5 file for the dataset](https://www.dropbox.com/s/brxao096sl5poxz/SPICE-corrected.hdf5.gz?dl=0).

I've tried training with or without derivatives.  I've tried a range of initial learning rates, with or without warmup.  I've tried varying model parameters.  I've tried restricting it to only molecules with no formal charges.  Nothing makes any difference.  In all cases, the loss starts out at about 2e7 and never decreases.

The dataset consists of all SPICE calculations that had been completed when I started working on this a couple of weeks ago.  I converted the units so positions are in Angstroms and energies in kJ/mol.  I also subtracted off per-atom energies.  Atom types are the union of element and formal charge.  Here's the mapping:

```python
typeDict = {('Br', -1): 0, ('Br', 0): 1, ('C', -1): 2, ('C', 0): 3, ('C', 1): 4, ('Ca', 2): 5, ('Cl', -1): 6,
            ('Cl', 0): 7, ('F', -1): 8, ('F', 0): 9, ('H', 0): 10, ('I', -1): 11, ('I', 0): 12, ('K', 1): 13,
            ('Li', 1): 14, ('Mg', 2): 15, ('N', -1): 16, ('N', 0): 17, ('N', 1): 18, ('Na', 1): 19, ('O', -1): 20,
            ('O', 0): 21, ('O', 1): 22, ('P', 0): 23, ('P', 1): 24, ('S', -1): 25, ('S', 0): 26, ('S', 1): 27}
```

If anyone can provide insight, I'll be very grateful!

```
activation: silu
atom_filter: -1
batch_size: 128
cutoff_lower: 0.0
cutoff_upper: 8.0
dataset: HDF5
dataset_root: SPICE-corrected.hdf5
derivative: false
distributed_backend: ddp
early_stopping_patience: 40
embedding_dimension: 64
energy_weight: 1.0
force_weight: 0.001
inference_batch_size: 128
lr: 1.e-4
lr_factor: 0.8
lr_min: 1.e-7
lr_patience: 10
lr_warmup_steps: 5000
max_num_neighbors: 80
max_z: 28
model: equivariant-transformer
neighbor_embedding: true
ngpus: -1
[...]

PR #84, opened 2022-05-12
Avoid multiple processes reading from the same HDF5 file

PhilippThoelke:
When using the HDF5 dataset class we sometimes run into multiprocessing issues where multiple workers read from the same h5py file instance, which might corrupt the data (discovered in #82). Reading from one h5py file instance using multiple processes is not recommended according to [their documentation](https://docs.h5py.org/en/stable/mpi.html). This PR avoids the problem by initializing the h5py files once in the dataset's `get` method instead of the constructor.

### DIFF
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -2,7 +2,10 @@ import pytest
 from pytest import mark, raises
 from os.path import join
 import numpy as np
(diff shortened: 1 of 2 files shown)


giadefa (approved):
I assume that you have tested it

PR #87, opened 2022-05-12
Fix CI

raimis:
- [x] Pin TorchANI package before the rebuild for PyTorch 1.11 (https://github.com/conda-forge/torchani-feedstock/pull/10/)


raimis:
@PhilippThoelke a quick fix. Later I'll update the rest dependencies.

PhilippThoelke (approved):

PR #88, opened 2022-05-12
see if we can use conda-forge NNPOps

mikemhenry:


raimis:
Yes, we should switch to NNPOps from `conda-forge`. I just did have to time to update.

raimis:
Alredy use NNPOps form `conda-forge` (#90)

PR #94, opened 2022-05-21
Handle when load_model is specified twice

PhilippThoelke:
Currently the config file overwrites the `load_model` argument from the command line, if specified. It would be more intuitive if the command line argument overwrites the config file as this matches the behavior of all other arguments. Now we show a warning if `load_model` is specified twice and the two arguments differ and we ignore `load_model` in the config file.


raimis (approved):
Looks good!

PR #97, opened 2022-06-01
Attempt at upgrading pytorch-lightning

PhilippThoelke:
We no longer seem to encounter problems with recent pytorch-lightning versions and jittable MessagePassing layers from pytorch-geometric. This PR upgrades the code base to be compatible with the changes made to the pytorch-lightning API from version 1.3.8 to 1.6.3.


stefdoerr (approved):
lgtm

scripts/train.py
```diff
@@ -45,7 +45,6 @@ def get_args():
     parser.add_argument('--test-interval', type=int, default=10, help='Test interval, one test per n epochs (default: 10)')
     parser.add_argument('--save-interval', type=int, default=10, help='Save interval, one save per n epochs (default: 10)')
     parser.add_argument('--seed', type=int, default=1, help='random seed (default: 1)')
-    parser.add_argument('--distributed-backend', default='ddp', help='Distributed backend: dp, ddp, ddp2')
```
giadefa, line 48: This would make it not compatible with previous runs. What is the way used now by pytorch lightning to assign the distributed backed?

peastman (approved):
Looks good.

tests/test_module.py
```diff
@@ -33,9 +33,10 @@ def test_train(model_name, use_atomref, tmpdir):
         test_size=None,
         log_dir=tmpdir,
         derivative=True,
-        embedding_dimension=32,
-        num_layers=3,
+        embedding_dimension=16,
```
raimis, line 36: Why do you change these defaults?

scripts/train.py
PhilippThoelke: The trainer class has `strategy` now, which is similar and can also take a training strategy string, like "ddp" for example, but some of possible training strategies changed in name and behavior so just by upgrading to new PL it will behave differently. If you think it's useful I can add the `--distributed-backend` argument back in and set it to a new, reasonable default that is passed to `strategy` in the trainer.

giadefa: These are the strategies:
https://pytorch-lightning.readthedocs.io/en/latest/extensions/strategy.html

I am more concerned that old scripts would directly fail without a why. Either we add a deprecated message or we add a default strategy.

PhilippThoelke: You mean using old config files, which include `distributed_backend`, would break? That would just throw an error saying that the argument doesn't exist.

I added a default strategy in the trainer: https://github.com/torchmd/torchmd-net/blob/136e5dd8b10e939bfdc6fe54951a9b1b76ee9d7b/scripts/train.py#L149
I can add the command line argument back in if you want but I would suggest changing the default to "ddp_find_unused_parameters_false".

giadefa:
But you world never use other strategies?

PhilippThoelke:
Most training strategies focus either on very large models (billion+ parameters), which I don't see happening here, or on very specific hardware (e.g. HPU, IPU or TPU). I think so far only using DDP has been working well.

giadefa (approved):

PR #98, opened 2022-06-03
reduce the dependencies of torchmd-net

stefdoerr:


torchmdnet/utils.py
```diff
@@ -5,6 +5,35 @@
 from os.path import dirname, join, exists
 from pytorch_lightning.utilities import rank_zero_warn
 
+# fmt: off
```
raimis, line 8: Maybe you could include a comment with the source of numbers.

raimis (approved):

torchmdnet/utils.py
stefdoerr: Sure, I took the comment on the masses now from ASE.

PR #102, opened 2022-07-05
QM9q: implement the reference energy computation

raimis:
- [x] Atomic energies
- [x] Reference energy computation for charged molecules

### DIFF
--- a/torchmdnet/datasets/qm9q.py
+++ b/torchmdnet/datasets/qm9q.py
@@ -61,6 +61,33 @@ class QM9q(Dataset):
 
         raise RuntimeError(f'Cannot load {paths}')
 
+    @staticmethod
+    def compute_reference_energy(atomic_numbers, charge):
+
+        atomic_numbers = np.array(atomic_numbers)
+        charge = int(charge)
+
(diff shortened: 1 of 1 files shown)


torchmdnet/datasets/qm9q.py
```diff
@@ -11,6 +11,26 @@ class QM9q(Dataset):
...
+             1: -53.9895574739},
+        8: {-1: -74.9699154500,
+             0: -74.9812632126,
+             1: -74.4776884006},
+        9: {-1: -99.6695561536,
+             0: -99.6185158728},
+    }
+
+    INITIAL_CHARGES = {element: sorted([(energy, charge) for charge, energy in charges.items()])[0][1]
+                       for element, charges in ATOM_ENERGIES.items()}
```
stefdoerr, line 32: Avoid double nested list comprehensions please. They are quite impossible to read

raimis: Split the loops and added more docs

raimis:
@stefdoerr can you finish reviewing?

stefdoerr (approved):
Looks good. I'm only curious if the charge atom energies are additive. Like does a +2 atom create equal energy increase as two +1 atoms?

PR #104, opened 2022-07-22
Fix SPICE loader

raimis:
- [x] Remove a workaround for https://github.com/openmm/spice-dataset/issues/30
- [x] Remove sample with large forces


stefdoerr:
I'm re-running the tests. Some weird error

raimis:
@stefdoerr is it the same error we talked?

stefdoerr:
No I meant here on the GH Actions. Seems like it passes now after merging main. Fine by me

stefdoerr (approved):

PR #105, opened 2022-07-22
CI: Use Mamba

raimis:
- [x] Use Mamba


stefdoerr (approved):
You can also waste some time of your life trying to make it work with micromamba if you have nothing better to do :P https://github.com/mamba-org/provision-with-micromamba

PR #106, opened 2022-07-26
Comp6 formation energies

stefdoerr:


stefdoerr:
You can see the change in the second commit. The first is just the formatter

raimis:
Don't impose `black` on a project which isn't using it.

torchmdnet/datasets/comp6.py
```diff
@@ -5,31 +5,50 @@
...
   - https://aip.scitation.org/doi/abs/10.1063/1.5023802
 
-'''
+"""
+
 
 class COMP6Base(Dataset):
 
-    ATOMIC_NUMBERS = {b'H': 1, b'C': 6, b'N': 7, b'O': 8}
+    ELEMENT_ENERGIES = {
```
raimis, line 21: Mention that this copied from ANI-1x.

torchmdnet/datasets/comp6.py
```diff
@@ -41,31 +60,49 @@ def raw_url_name(self):
...
+                    dtype=pt.long,
+                )
+                all_pos = pt.tensor(mol["coordinates"][:], dtype=pt.float32)
+                all_y = pt.tensor(
+                    mol["energies"][:] * self.HARTREE_TO_EV, dtype=pt.float64
+                )
+                all_dy = pt.tensor(
+                    mol["forces"][:] * self.HARTREE_TO_EV, dtype=pt.float32
+                )
+                all_y -= self.compute_reference_energy(z)
```
raimis, line 105: Probably, we need to change other loader. Otherwise, it just shifts the inconsistency.

stefdoerr:
Ok, I'll create a separate black PR for the whole repo then

torchmdnet/datasets/comp6.py
stefdoerr: Which? QM9q does this now and SPICE returns formation energies. We should implement it for ANI-1x as well I guess

PR #109, opened 2022-07-28
auto-formatting with black

stefdoerr:
nothing else was done
Just to keep consistent style between all our code-bases


raimis (approved):
OK! I truss you that `black` hasn't changed the code.

PR #110, opened 2022-07-28
Bug? MD17 stores all raw paths to the same processed_path

basveeling:
Hi!

I might be mistaken here, but it seems there is a small bug in the MD17 dataloader. If so, I think this would fix it.

Thanks for making TorchMDNet open source :-)!


raimis (approved):
Thanks for catching the bug.

PR #112, opened 2022-08-02
Update the installation instructions

### LINKED ISSUE #108: Problem creating an environment
I install torchmd-net as shown in README.md. When I run examples with the command:
`CUDA_VISIBLE_DEVICES=0 python scripts/train.py --conf examples/ET-QM9.yaml`,
 it reports the following errors:

d1bbdee55949643f6c982c56525685d8

Can anyone know the reasons?

raimis:
- [x] Update the installation instructions to circumvent https://github.com/conda-forge/openmm-torch-feedstock/issues/20

Fixes #108

### DIFF
--- a/README.md
+++ b/README.md
@@ -4,22 +4,24 @@ TorchMD-NET provides state-of-the-art graph neural networks and equivariant tran
 
 ## Installation
 
-Clone the repository:
-```
-git clone https://github.com/torchmd/torchmd-net.git
-cd torchmd-net
-```
+1. Clone the repository:
+    ```
+    git clone https://github.com/torchmd/torchmd-net.git
+    cd torchmd-net
+    ```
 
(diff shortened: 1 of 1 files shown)


raimis:
@stefdoerr could you review?

stefdoerr:
The only difference I've seen between mamba and conda is that mamba will ignore channel priority if it finds higher package versions in later channels. Are you sure this is not the case here as well? Because then it could be solved by asking for the right package versions

raimis:
No, it is an issue like in https://github.com/conda-forge/openmm-torch-feedstock/issues/20: `conda` installs an incompatible package with from the same channel.

raimis:
The package priorities isn't an issue, because everything comes from `conda-forge`.

stefdoerr (approved):

PR #113, opened 2022-08-19
Improved SPICE loader

raimis:
- [x] Automatically download the dataset
- [x] Allow to select subsets
- [x] Allow to set gradient filter

### DIFF
--- a/torchmdnet/datasets/spice.py
+++ b/torchmdnet/datasets/spice.py
@@ -1,4 +1,4 @@
-from base64 import b64encode
+import hashlib
 import h5py
 import numpy as np
 import os
@@ -42,8 +42,8 @@ class SPICE(Dataset):
         pre_filter=None,
         dataset_arg=None,
     ):
-        self.dataset_arg = str(dataset_arg)
-        dataset_arg_hash = b64encode(self.dataset_arg.encode()).decode()
+        self.dataset_arg = "{}" if dataset_arg is None else str(dataset_arg)
+        dataset_arg_hash = hashlib.md5(self.dataset_arg.encode()).hexdigest()
         self.name = f"{self.__class__.__name__}-{dataset_arg_hash}"
         super().__init__(root, transform, pre_transform, pre_filter)
 
@@ -65,8 +65,6 @@ class SPICE(Dataset):
     def sample_iter(self):
 
         assert len(self.raw_paths) == 1
-
-        print(self.dataset_arg)
 
         for mol in tqdm(h5py.File(self.raw_paths[0]).values(), desc="Molecules"):
 
--- a/torchmdnet/datasets/spice.py
+++ b/torchmdnet/datasets/spice.py
@@ -16,7 +16,7 @@ class SPICE(Dataset):
     The subsets can be selected with `subsets`. By default, all the subsets are loaded.
 
     For example, this loads just two subsets:
-    >>> ds = ds = SPICE(".", subsets=["SPICE PubChem Set 1 Single Points Dataset v1.2", "SPICE PubChem Set 2 Single Points Dataset v1.2"])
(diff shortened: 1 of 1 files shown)


raimis:
@stefdoerr could you review?

torchmdnet/datasets/spice.py
```diff
@@ -1,21 +1,50 @@
...
 
 
 class SPICE(Dataset):
 
     """
     SPICE dataset (https://github.com/openmm/spice-dataset)
+
+    The dataset consist of several subsets (https://github.com/openmm/spice-dataset/blob/main/downloader/config.yaml). By default all the subsets are loaded or they can be selected with `dataset_arg`. For example, this loads just two subsets:
+
+    >>> ds = SPICE('.', dataset_arg='{"subsets": ["SPICE PubChem Set 1 Single Points Dataset v1.2", 'SPICE PubChem Set 2 Single Points Dataset v1.2']}')
```
stefdoerr, line 18: Why pass a string of a json instead of a python dict?

raimis: Because the argument parser doesn't support nested dictionaries: https://github.com/torchmd/torchmd-net/blob/080cf2214b146f8c92230d34dc3435136e47a604/scripts/train.py#L20-L92

stefdoerr: I would still make the class accept python dicts. Just `json.parse` the string of argparse if you are using command line to convert it to python dict and pass that to the class. Makes the python API layer more usable

raimis: Other loaders expect `loader_arg` to be `str`. How should I handle that?

stefdoerr: I looked at the other loaders. They accept single strings of undocumented parameters so I don't feel like it's a standard. Like here where it's silently converted to `paths` https://github.com/torchmd/torchmd-net/blob/main/torchmdnet/datasets/qm9q.py#L35-L61. I would argue that the CLI should always accept a json string even for the old datasets (like `'{"paths": "/tmp/"}'` and convert it to python dicts and the old datasets should be modified to read the specific key/value from the dict. That makes it also clearer what arguments are being set from command line.

raimis: @stefdoerr Yes, it would be a good solution. Could you make a PR with the proposed changes for the old loaders? After I'll follow your lead with this PR.

torchmdnet/datasets/spice.py
```diff
@@ -12,15 +12,17 @@ class SPICE(Dataset):
     """
     SPICE dataset (https://github.com/openmm/spice-dataset)
 
-    The dataset consist of several subsets (https://github.com/openmm/spice-dataset/blob/main/downloader/config.yaml). By default all the subsets are loaded or they can be selected with `dataset_arg`. For example, this loads just two subsets:
+    The dataset consists of several subsets (https://github.com/openmm/spice-dataset/blob/main/downloader/config.yaml).
+    The subsets can be selected with `subsets`. By default, all the subsets are loaded.
 
-    >>> ds = SPICE('.', dataset_arg='{"subsets": ["SPICE PubChem Set 1 Single Points Dataset v1.2", 'SPICE PubChem Set 2 Single Points Dataset v1.2']}')
+    For example, this loads just two subsets:
+    >>> ds = ds = SPICE(".", subsets=["SPICE PubChem Set 1 Single Points Dataset v1.2", "SPICE PubChem Set 2 Single Points Dataset v1.2"])
```
stefdoerr, line 19: ds = ds =
typo

raimis: Good catch!

raimis:
@stefdoerr could you review again?

stefdoerr (approved):
lgtm

PR #114, opened 2022-08-22
QM9q: load dipoles and partial charges

raimis:
- [x] Load dipole moments
- [x] Load partial charges


raimis:
@stefdoerr could you review?

stefdoerr (approved):

PR #115, opened 2022-08-23
Refactor the additional argument passing to dataset classes

stefdoerr:
This cleans up the logic of passing extra datasets arguments to dataset classes.

### DIFF
--- a/torchmdnet/datasets/md17.py
+++ b/torchmdnet/datasets/md17.py
@@ -36,6 +36,10 @@ class MD17(InMemoryDataset):
             molecules = ",".join(MD17.available_molecules)
         self.molecules = molecules.split(",")
 
+        for mol in molecules:
+            if mol not in MD17.available_molecules:
+                raise RuntimeError(f"Molecule '{mol}' does not exist in MD17")
+
         if len(self.molecules) > 1:
             rank_zero_warn(
                 "MD17 molecules have different reference energies, "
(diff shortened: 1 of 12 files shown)


torchmdnet/datasets/md17.py
```diff
@@ -25,16 +25,16 @@ class MD17(InMemoryDataset):
 
     available_molecules = list(molecule_files.keys())
 
-    def __init__(self, root, transform=None, pre_transform=None, dataset_arg=None):
-        assert dataset_arg is not None, (
+    def __init__(self, root, transform=None, pre_transform=None, molecules=None):
+        assert molecules is not None, (
```
raimis, line 29: Maybe it is better to check:
```python
assert molecules in MD17.available_molecules or molecules == "all"
```

torchmdnet/datasets/qm9.py
```diff
@@ -5,14 +5,14 @@
 
 
 class QM9(QM9_geometric):
-    def __init__(self, root, transform=None, dataset_arg=None):
-        assert dataset_arg is not None, (
+    def __init__(self, root, transform=None, label=None):
+        assert label is not None, (
```
raimis, line 9: Maybe it is better to check:
```python
assert label in qm9_target_dict.values()
```

raimis (commented):
This look good in general. Just some argument checks can be more strict as commented.

torchmdnet/datasets/md17.py
stefdoerr: Done, in a bit different way

torchmdnet/datasets/qm9.py
stefdoerr: Done

raimis (approved):

PR #118, opened 2022-08-26
Fix the usage of gradients

### LINKED ISSUE #116: Inconsistency between gradients and forces
When training with forces, the model computes the negative gradient of the energy with respect to the positions (a.k.a. forces):
https://github.com/torchmd/torchmd-net/blob/db72e12461b1bce9bb3ddebd093fa11a803d37ca/torchmdnet/models/model.py#L196-L207

Some the loaders load forces:
https://github.com/torchmd/torchmd-net/blob/db72e12461b1bce9bb3ddebd093fa11a803d37ca/torchmdnet/datasets/ani.py#L279-L281
https://github.com/torchmd/torchmd-net/blob/db72e12461b1bce9bb3ddebd093fa11a803d37ca/torchmdnet/datasets/comp6.py#L100-L102

While the other ones load gradients (the opposite sign):
https://github.com/torchmd/torchmd-net/blob/db72e12461b1bce9bb3ddebd093fa11a803d37ca/torchmdnet/datasets/qm9q.py#L147-L151
https://github.com/torchmd/torchmd-net/blob/db72e12461b1bce9bb3ddebd093fa11a803d37ca/torchmdnet/datasets/spice.py#L99-L103

We need to agree what `y` represents.

raimis:
- [x] Fix the gradient calculation
- [x] Update the loaders

Fixes #116

### DIFF
--- a/torchmdnet/datasets/ace.py
+++ b/torchmdnet/datasets/ace.py
@@ -34,7 +34,7 @@ class Ace(Dataset):
             z=pt.tensor(mol["atomic_numbers"], dtype=pt.long),
             pos=pt.tensor(mol["positions"][iconf], dtype=pt.float32),
             y=pt.tensor(mol["energy"][iconf], dtype=pt.float32),
-            dy=pt.tensor(mol["forces"][iconf], dtype=pt.float32),
+            dy=-pt.tensor(mol["forces"][iconf], dtype=pt.float32),
             d=pt.tensor(mol["dipole_moment"][iconf], dtype=pt.float32),
         )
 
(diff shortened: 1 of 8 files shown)


PhilippThoelke:
This will potentially break simulations. If you want to change model outputs to gradients instead of forces, you also have to adjust e.g. the torchMD calculator to then use -derivative. I find it more intuitive though if the model returns forces as that's what we actually care about.

giadefa:
Yes, the model must return forces

raimis:
The issue wasn't fixed.

giadefa:
If you refer to the confusing naming, Philipp will make a PR it next week.

For the sign, it was fixed by your PR.

PR #119, opened 2022-08-26
Fix the gradient sign for QM9q and SPICE

raimis:
A quick fix for #116 until a proper solution decided.

### DIFF
--- a/torchmdnet/datasets/spice.py
+++ b/torchmdnet/datasets/spice.py
@@ -97,7 +97,7 @@ class SPICE(Dataset):
                 * self.HARTREE_TO_EV
             )
             all_dy = (
-                pt.tensor(mol["dft_total_gradient"], dtype=pt.float32)
+                -pt.tensor(mol["dft_total_gradient"], dtype=pt.float32)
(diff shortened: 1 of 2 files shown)


raimis:
@stefdoerr could you review?

giadefa (approved):

torchmdnet/datasets/spice.py
```diff
@@ -97,7 +97,7 @@ def sample_iter(self):
                 * self.HARTREE_TO_EV
             )
             all_dy = (
-                pt.tensor(mol["dft_total_gradient"], dtype=pt.float32)
+                -pt.tensor(mol["dft_total_gradient"], dtype=pt.float32)
```
giadefa, line 100: From the HDF5 loader i understand that there is a group["forces"] in the SPICE dataset, so why you use dft_total_gradient?

peastman: They're totally different formats.  They just happen to both use HDF5.

stefdoerr (approved):

PR #128, opened 2022-09-29
Implement molecule sub-sampling

raimis:
- [x] Implement molecule sub-sampling
  - [x] Ace
  - [x] SPICE
- [x] Update docs


stefdoerr (approved):
It's ok but it depends on the order of the molecules which can produce very weird subsets.

PR #133, opened 2022-10-06
Move the reduction into the output model

raimis:
In order to predict per-atom properties (e.g. partial charges), the reduction operation has to be redefined. Currently, the reduction operations are limited to the ones supported by `scatter` (https://github.com/rusty1s/pytorch_scatter).

This PR make the reduction operation parts of the output model and allows the model-specific implementations.

### DIFF
--- a/torchmdnet/models/output_modules.py
+++ b/torchmdnet/models/output_modules.py
@@ -1,4 +1,5 @@
 from abc import abstractmethod, ABCMeta
+from torch_scatter import scatter
 from typing import Optional
 from torchmdnet.models.utils import act_class_mapping, GatedEquivariantBlock
 from torchmdnet.utils import atomic_masses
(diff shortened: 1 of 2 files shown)


raimis:
@stefdoerr @PhilippThoelke could you review?

torchmdnet/models/output_modules.py
```diff
@@ -11,9 +12,10 @@
 
 
 class OutputModel(nn.Module, metaclass=ABCMeta):
-    def __init__(self, allow_prior_model):
+    def __init__(self, allow_prior_model, reduce_op):
```
stefdoerr, line 15: Old one had a default for reduce_op="add". Not sure if it's still needed though. Do all outputmodels explicitly define their reduction operation?

stefdoerr (commented):
Much appreciated! I need this a while ago myself but didn't get to implementing it.

stefdoerr (approved):

Issue #26, opened 2021-06-23
Physics based priors


peastman:
I want to implement some physics based priors.  An example would be a Coulomb interaction based on pre-computed partial charges that are stored in the dataset.  `BasePrior.forward()` is supposed to return a list of per-atom energy contributions, but physics based interactions usually do not decompose in that way.  It would be much easier if it could just return a total energy for each sample.

What do you recommend as the cleanest way of implementing this?

peastman:
Sorry for reopening an old issue - this never actually got implemented!  I'm about to start on it now.  I'll add the `atomwise_prior` flag as suggested above.  I also have a few other questions about the best way of implementing particular potentials.

Physics based potentials will often require extra information.  For example, one prior I want to add is the [ZBL stopping potential](https://en.wikipedia.org/wiki/Stopping_power_(particle_radiation)#Repulsive_interatomic_potentials) for short range repulsion.  It requires knowing the atomic number of every atom.  `forward()` takes an argument `z` which often corresponds to the atomic number, but not always.  For example, in some of my models I define atom types by the combination of element and formal charge, so `z` is an arbitrary atom type index.  I need to separately provide both an atom type and an atomic number.

A more complicated example is Coulomb, which depends on the charge of each atom.  That could be a fixed number (formal charge or precomputed partial charge), but in other cases it will itself be calculated as an output of the model.

What's the best way of providing extra information like this that's needed to compute a physics based potential?

peastman:
Another question: currently it only lets you specify a single prior model, but often you may want more than one.  A ZBL potential for short range repulsion, plus Coulomb for long range interactions, plus D4 for dispersion.  (SpookyNet includes all of those.)  Any thoughts on the best way of handling this?

peastman:
And another question: how do we want to handle units?  Nothing in the current code worries about them, except sometimes for a particular dataset.  The inputs and outputs to a model are just numbers that might have any units.  But physical potentials involve hardcoded constants like `eps_0` whose values depend on units.  How should we handle this?  Some possibilities include

- Standardizing on a particular set of units for all models.
- Standardize on a particular set of units for a particular potential function, and require the user to provide scaling factors for converting positions, energies, etc. to the required units.
- Implement it at a higher level, where a model specifies what units it uses.
  - We could also make each Dataset specify its units, and conversions between the dataset and model units would happen automatically.

giadefa:
can you use that single model as a wrapper for all the models that you
[...]

PR #134, opened 2022-10-06
Implement ZBL potential

peastman:
This is the first piece of #26.  It isn't fully tested yet, but I think it's ready for a first round of comments.

I created a mechanism for passing arguments to the prior.  `prior_args` is now the option specified by the user in the config file.  `prior_init_args` stores the value returned by `get_init_args()`, which contains the arguments needed for reconstructing it from a checkpoint.

This prior requires the dataset to provide several pieces of information.  To keep things general, the HDF5 format allows the file to contain a `_metadata` group which can store arbitrary pieces of information.  Most of the other dataset classes should be able to hardcode the necessary values, since they aren't intended to be general.


stefdoerr:
Looks good to me. Tests are failing though:
```
 E       RuntimeError: 
E       'NoneType' object has no attribute or method 'atomwise'.:
E         File "/usr/share/miniconda/envs/torchmd-net/lib/python3.9/site-packages/torchmdnet/models/model.py", line 182
E           
E               # apply atomwise prior model
E               if self.prior_model is not None and self.prior_model.atomwise:
E                                                   ~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
E                   x = self.prior_model(x, z, pos, batch)
```

Also you might have some merge conflicts with the latest PR of Raimondas

stefdoerr:
Wait a second, how does this error even occur? Shouldn't the `if` short-circuit if self.prior_model is not None? Weird

torchmdnet/models/model.py
```diff
@@ -178,13 +178,17 @@ def forward(
         if self.std is not None:
             x = x * self.std
 
-        # apply prior model
-        if self.prior_model is not None:
+        # apply atomwise prior model
```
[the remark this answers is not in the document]
peastman: I can go either way on that.  I was following what @PhilippThoelke recommended in https://github.com/torchmd/torchmd-net/issues/26#issuecomment-867448271.

torchmdnet/models/model.py
```diff
@@ -178,13 +178,17 @@ def forward(
...
+        if self.prior_model is not None and self.prior_model.atomwise:
             x = self.prior_model(x, z, pos, batch)
 
         # aggregate atoms
         out = scatter(x, batch, dim=0, reduce=self.reduce_op)
 
+        # apply non-atomwise prior model
+        if self.prior_model is not None and not self.prior_model.atomwise:
+            out += self.prior_model(x, z, pos, batch)
```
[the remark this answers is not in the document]
peastman: What do you suggest to make it more general?  What other types of priors would be useful?

torchmdnet/priors.py
```diff
@@ -76,3 +79,52 @@ def get_init_args(self):
...
+class ZBL(BasePrior):
+    """This class implements the Ziegler-Biersack-Littmark (ZBL) potential for screened nuclear repulsion.
+    Is is described in https://doi.org/10.1007/978-3-642-68779-2_5 (equations 9 and 10 on page 147).  It
+    is an empirical potential that does a good job of describing the repulsion between atoms at very short
+    distances.
+
+    To use this prior, the Dataset must provide the following attributes.
+
+    atomic_number: 1D tensor of length max_z.  atomic_number[z] is the atomic number of atoms with atom type z.
+    distance_scale: multiply by this factor to convert coordinates stored in the dataset to meters
```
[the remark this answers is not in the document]
peastman: I went with SI units to avoid preferencing any particular choice of units.  For what it's worth, eV is rarely used in MD simulations.  Most MD codes use either kJ/mol or kcal/mol as their energy unit.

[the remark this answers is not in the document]
giadefa: Can't you do the opposite? I.e. specify in the ZBL prior what units are used by the NNP?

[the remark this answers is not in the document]
giadefa: This way everyone can keep using their units and ZBL will adapt to it.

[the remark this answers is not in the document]
peastman: The neural network is agnostic with regard to units.  It receives inputs and tries to make its outputs match the ones in the dataset.  It doesn't know or care what units they're in.

A physics based potential does depend on units.  It needs to know what units the coordinates are in, and it needs to know what units the energy is expected to be in.

giadefa:
Who runs the model knows what units want to use and set it for the priors

peastman:
> Who runs the model knows what units want to use and set it for the priors

The units are determined by the dataset.  It isn't a free choice for the user to make.  If your dataset contains positions in nm and energies in kJ/mol, that's what the prior needs to work with.  Any other units would produce wrong results.

We could consider creating an automated unit conversion system like I suggested in https://github.com/torchmd/torchmd-net/issues/26#issuecomment-1266154668, but that's a separate project.

peastman:
> Wait a second, how does this error even occur? Shouldn't the if short-circuit if self.prior_model is not None?

Yes it should!  I think it's a bug in the torchscript jit.  I was able to work around it by splitting it into two lines.

peastman:
This is ready for review.

stefdoerr:
The test is failing if you could take a look

torchmdnet/scripts/train.py
```diff
@@ -63,6 +63,7 @@ def get_args():
     parser.add_argument('--model', type=str, default='graph-network', choices=models.__all__, help='Which model to train')
     parser.add_argument('--output-model', type=str, default='Scalar', choices=output_modules.__all__, help='The type of output model')
     parser.add_argument('--prior-model', type=str, default=None, choices=priors.__all__, help='Which prior model to use')
+    parser.add_argument('--prior-args', default=None, type=str, help='Additional arguments for the prior model. Need to be specified in JSON format i.e. \'{"cutoff_distance": 10.0, "max_num_neighbors": 100}\'')
```
stefdoerr, line 66: how does this work? I don't see the json being converted to a dictionary anywhere. Also how would it work from a yaml config file? could you provide some examples?

stefdoerr (changes requested):

torchmdnet/scripts/train.py
peastman: Although people mostly use yaml for simple config files, it actually is a strict superset of json.  So there's nothing to convert: json syntax is yaml syntax, and the dict gets parsed automatically.

peastman:
> The test is failing if you could take a look

Fixed.  I couldn't reproduce it locally, but I managed to guess what it was unhappy about.

stefdoerr (approved):
Fine by me. Does anyone else want to comment?

tests/test_module.py
```diff
@@ -43,7 +43,7 @@ def test_train(model_name, use_atomref, tmpdir):
     prior = None
     if use_atomref:
         prior = getattr(priors, args["prior_model"])(dataset=datamodule.dataset)
-        args["prior_args"] = prior.get_init_args()
+        args["prior_init_args"] = prior.get_init_args()
```
raimis, line 46: The loading of the pretrained models (https://github.com/torchmd/torchmd-net/tree/main/examples#loading-checkpoints) fails:

```python
from torchmdnet.models.model import load_model
load_model('ANI1-equivariant_transformer/epoch=359-val_loss=0.0004-test_loss=0.0120.ckpt')

AssertionError: Requested prior model Atomref but the arguments are lacking the key "prior_init_args".
```

raimis (changes requested):

tests/test_module.py
peastman: It sounds like we want to redo how prior args are specified as described in https://github.com/torchmd/torchmd-net/issues/26#issuecomment-1299237398.  That means we can switch back to `prior_args` for this value.  There will still be compatibility issues, because it will need to become a list with args for multiple prior models, but I can add a check for that case for backward compatibility.  I'll go ahead and make the changes in this PR.

PhilippThoelke: Might be a good idea to add a test case for loading model checkpoints from a previous version.

torchmdnet/priors/zbl.py
```diff
@@ -0,0 +1,54 @@
...
+    atomic_number: 1D tensor of length max_z.  atomic_number[z] is the atomic number of atoms with atom type z.
+    distance_scale: multiply by this factor to convert coordinates stored in the dataset to meters
+    energy_scale: multiply by this factor to convert energies stored in the dataset to Joules (*not* J/mol)
+    """
+    def __init__(self, cutoff_distance, max_num_neighbors, atomic_number=None, distance_scale=None, energy_scale=None, dataset=None):
+        super(ZBL, self).__init__()
+        if atomic_number is None:
+            atomic_number = dataset.atomic_number
+        if distance_scale is None:
+            distance_scale = dataset.distance_scale
```
raimis, line 22: The dataset have atom positions, not distance. So, for consistency, `distance_scale` should be `position_scale`.

peastman:
I added the multiple priors support.  I tried to make the syntax fairly flexible.  All of the following are valid.

```yaml
prior_model: Atomref
```

```yaml
prior_model:
  - ZBL:
      cutoff_distance: 4.0
      max_num_neighbors: 50
  - Atomref
```

```yaml
prior_model:
  - ZBL:
      cutoff_distance: 4.0
      max_num_neighbors: 50
    Atomref:
```

I added tests for creating models both from config files and from checkpoints.  And the test for older checkpoints in test_module.py continues to work.  I can't be sure I've caught all possible problems though, so if you can check your own files that would be helpful.

torchmdnet/priors/zbl.py
peastman: The dataset also contains forces (units of energy/distance).  The scale factor gets applied to both.  And technically speaking, meters are a unit of distance, not position.  A position is a compound quantity made up of a reference frame and a displacement vector measured in that reference frame.

peastman:
Is this ok to merge?

raimis (approved):
It is good to go!

PR #140, opened 2022-10-19
Allow the Ace loader to load multiple dataset at once

raimis:
- [x] Allow the Ace loader to load multiple dataset at once


stefdoerr (approved):
If somebody passes a tuple it will fail but otherwise ok. In moleculekit I wrote a `ensurelist` function which makes sure that whatever it is (tuple, list, np.ndarray) it will be converted to list.

PR #141, opened 2022-10-20
POC: Enable to train on the partial charges

raimis:
DO NOT MERGE!

- [x] Enable to train on the partial charges


raimis:
This is obsolted

PR #142, opened 2022-10-25
Support atom-wise and molecule-wise priors

raimis:
Implement the support of atom-wise (e.g. atomic reference energies) and molecule-wise (e.g. electrostatic term) priors.

- [x] Refactor the priors
- [x] Implement pre-reduce and post-reduce methods


peastman:
This conflicts with #134, which implements the same feature in a different way.  Can we agree on which API we want to use?  I followed what @PhilippThoelke suggested.  If we decide to use the other API, let me change it there so we don't create conflicts.

raimis:
@stefdoerr @PhilippThoelke what do you think?

PhilippThoelke:
I think both #134 and this PR will be useful and as far as I understand it they're not mutually exclusive? @peastman or is the conflict more substantial?

peastman:
> I think both https://github.com/torchmd/torchmd-net/pull/134 and this PR will be useful and as far as I understand it they're not mutually exclusive?

We just need to pick which API we prefer for the priors.  Either one can work for my needs.  I don't think it would make sense to do both, since they're basically alternate APIs for doing the same thing.

PhilippThoelke:
I would say the API in this PR is more general and therefore preferable but I think it's still useful to keep e.g. the distinction between `prior_args` and `prior_init_args` from #134.

torchmdnet/priors/atomref.py
```diff
@@ -1,54 +1,17 @@
-from abc import abstractmethod, ABCMeta
+from .base import BasePrior
```
stefdoerr, line 1: Use absolute instead of local imports. They mess up some testing frameworks

stefdoerr:
I agree with Philipp, this implementation seems more generic but should be trivial to merge with the additional changes in #134

stefdoerr (approved):

PR #143, opened 2022-10-26
Prior: D2 dispersive correction

raimis:
- [x] Implement the D2 dispersive correction
- [x] Tests
- [x] Documentation


peastman:
This will be a really useful feature to have.

Can you use the same system as in #134, where the dataset provides `atomic_number`, `distance_scale`, and `energy_scale` fields?  That will allow it to be used with arbitrary units, and with datasets where atom types are different from atomic numbers.

tests/test_prior_d2.py
```diff
@@ -0,0 +1,83 @@
+from torchmdnet.priors import D2
+import torch as pt
+from pytest import mark
+
+
+def compute_with_psi4(z, pos):
```
peastman, line 6: This doesn't seem to be used anywhere.  Is it just to document how you computed the values below?  Maybe add a comment saying that?

torchmdnet/priors/__init__.py
```diff
@@ -1 +1,4 @@
 from torchmdnet.priors.atomref import Atomref
+from torchmdnet.priors.d2 import D2
+
+__all__ = ["Atomref", "D2"]
```
stefdoerr, line 4: why do we use __all__? I always found it sort of useless

stefdoerr (commented):
Can't help much as I don't really know what this dispersive correction is. I leave it to Peter to review

torchmdnet/priors/__init__.py
peastman: In this particular case, it gets used by the training script:

https://github.com/torchmd/torchmd-net/blob/9e0fef3f4f52ec6a40326d54cc71fd7675eb2ac9/torchmdnet/scripts/train.py#L66

It doesn't have to be implemented that way.  It just needs to get the list of allowed priors from somewhere.

stefdoerr: I see ok. Thanks!

torchmdnet/priors/d2.py
```diff
@@ -0,0 +1,187 @@
...
+        if ij.shape[1] == 0:
+            return y
+
+        # Compute the pair parameters
+        Z = self.Z_map[z[ij]]
+        C_6 = self.C_6[Z].prod(dim=0).sqrt()
+        R_r = self.R_r[Z].sum(dim=0)
+
+        # Compute pair contributions
+        f_dump = 1 / (1 + pt.exp(-self.d * (R_ij / R_r - 1)))
```
peastman, line 178: I think you meant to call this `f_damp` rather than `f_dump`?  It's a damping function.

tests/test_prior_d2.py
raimis: Yes, add a comment.

tests/test_prior_d2.py
```diff
@@ -0,0 +1,85 @@
...
+    energy = disp.compute_energy(mol)
+
+    return energy
+
+
+# The values of `y` are computed with the function above.
+# NOTE: Psi4 is not compatible with conda-forge
+TEST_CASES = {
+    "he": {
+        "z": [1],
```
peastman, line 35: If this is He, shouldn't z be 2?

torchmdnet/priors/d2.py
```diff
@@ -0,0 +1,187 @@
...
+            [31.74, 1.892],  # 52 Te
+            [31.50, 1.892],  # 53 I
+            [29.99, 1.881],  # 54 Xe
+        ]
+    )
+    C_6_R_r[:, 1] *= 0.1  # Å --> nm
+
+    def __init__(
+        self,
+        cutoff,
```
peastman, line 112: Can we call this `cutoff_distance` to be consistent with ZBL?

raimis: Done!

tests/test_prior_d2.py
raimis: Done!

raimis:
@peastman any more comments?

torchmdnet/priors/d2.py
```diff
@@ -0,0 +1,187 @@
...
+    """Dispersive correction term as used in DFT-D2
+
+    Reference
+    ---------
+    Grimme, Stefan. "Semiempirical GGA-type density functional constructed with a long‐range dispersion correction." Journal of computational chemistry 27.15 (2006): 1787-1799.
+    https://onlinelibrary.wiley.com/doi/10.1002/jcc.20495
+
+    Arguments
+    ---------
+    cutoff: float
```
peastman, line 17: You missed changing the name here.

peastman (approved):
Looks good!  Aside from the one last change to the docstring, I think it's ready to merge.

torchmdnet/priors/d2.py
raimis: Done!

Issue #148, opened 2022-11-06
Torchmd-net training crashed on Perlmutter GPU


FranklinHu1:
Hello,

I have been trying to train the torchmd-net model on the DOE NERSC Perlmutter system which is focused on GPU-accelerated applications. I am running into a strange error when training where the training will crash when it is time to test the model, with an index out of bounds error for one of the data loaders being the issue. 

# Installation 

I installed torchmd-net into my Perlmutter environment following the instructions given in the README. Before doing any training with the model, I am always sure to activate the torchmd-net environment with 

`% mamba activate torchmd-net`

My environment is as follows:

name: torchmd-net

channels:
  - conda-forge
  - defaults
 
dependencies:
  - _libgcc_mutex=0.1=conda_forge
  - _openmp_mutex=4.5=2_kmp_llvm
  - absl-py=1.2.0=pyhd8ed1ab_0
  - aiohttp=3.8.3=py39hb9d737c_0
  - aiosignal=1.2.0=pyhd8ed1ab_0
  - alsa-lib=1.2.7.2=h166bdaf_0
  - ambertools=22.0=py39hcb9d944_3
  - arpack=3.7.0=hdefa2d7_2
  - astunparse=1.6.3=pyhd8ed1ab_0
  - async-timeout=4.0.2=py39h06a4308_0
  - attr=2.5.1=h166bdaf_1
  - attrs=22.1.0=pyh71513ae_1
  - blinker=1.4=py39h06a4308_0
  - blosc=1.21.1=h83bc5f7_3
  - boost=1.74.0=py39h5472131_5
  - boost-cpp=1.74.0=h75c5d50_8
  - brotli=1.0.9=h5eee18b_7
  - brotli-bin=1.0.9=h5eee18b_7
  - brotlipy=0.7.0=py39h27cfd23_1003
  - bzip2=1.0.8=h7b6447c_0
  - c-ares=1.18.1=h7f8727e_0
  - ca-certificates=2022.9.24=ha878542_0
  - cached-property=1.5.2=hd8ed1ab_1
  - cached_property=1.5.2=pyha770c72_1
  - cachetools=5.2.0=pyhd8ed1ab_0
  - cairo=1.16.0=ha61ee94_1014
  - certifi=2022.9.24=pyhd8ed1ab_0
  - cffi=1.15.1=py39h74dc2b5_0
  - cftime=1.6.2=py39h2ae25f5_1
[...]

PR #149, opened 2022-11-06
Add missing `__all__` attribute to priors

PhilippThoelke:
The `__all__` attribute got lost in a recent commit, readding it here. It lists all implemented prior models and is used when train.py is called with `--help`.

This problem was mentioned in #148.


peastman:
The fix for this is already in #134.  If you're ok with the current state of it, we can just merge that one.  Otherwise, this will create a conflict for it.

PhilippThoelke:
Ah I didn't realize it was fixed there, closing this in favor of #134.

PR #156, opened 2022-11-28
Don't store tensors in init args

peastman:
On some computers (but not others) this leads to an error when resuming training, due to tensors stored in `hparams.yaml`.  I'm not sure why it only happens on some computers.

### DIFF
--- a/torchmdnet/priors/zbl.py
+++ b/torchmdnet/priors/zbl.py
@@ -1,5 +1,6 @@
 import torch
 from torchmdnet.priors.base import BasePrior
+from torchmdnet.priors.d2 import D2
 from torchmdnet.models.utils import Distance, CosineCutoff
 
 class ZBL(BasePrior):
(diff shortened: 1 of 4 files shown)


PhilippThoelke:
Any reason you used the `.numpy()` list comprehension and `float()` over `tensor.tolist()` and `tensor.item()`? The latter are safer when tensors are on different devices.

peastman:
Good idea.  I just changed it.

peastman:
All tests are now passing.

I also fixed a second error that prevented resuming training when using D2.  The constructor received the scale factors from arguments or from the dataset, then converted them to different units.  That meant `get_init_args()` would return the converted values rather than the ones that were actually passed in.  When you tried to resume training, it would therefore receive the wrong values, which lead to an exception.

torchmdnet/priors/zbl.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, cutoff_distance, max_num_neighbors, atomic_number=None, dista
             distance_scale = dataset.distance_scale
         if energy_scale is None:
             energy_scale = dataset.energy_scale
-        atomic_number = torch.as_tensor(atomic_number, dtype=torch.int8)
+        atomic_number = torch.as_tensor(atomic_number, dtype=torch.long)
```
stefdoerr, line 25: is there a reason we need to store atomic numbers as int64? Seems excessive

stefdoerr (commented):
Beyond my comment on the torch.long, it looks good to me

torchmdnet/priors/zbl.py
peastman: Using int8 led to exceptions on the GPU.  This is just a single array whose length equals the number of atom types (generally less than 100).  The storage is negligible.

peastman:
> Beyond my comment on the torch.long, it looks good to me

Can you submit a review approving it?  I can't merge it until you do.

stefdoerr (approved):

peastman:
Thanks!

PR #158, opened 2022-12-03
Allow datasets to pass extra fields to models

peastman:
This implements the feature described in https://github.com/torchmd/torchmd-net/issues/26#issuecomment-1334583040.  The dataset can store arbitrary extra fields in the Data objects it returns.  They get passed to the model in the `extra_args` parameter.  At the moment, the model only passes them on to the priors, since that's the initial thing I need them for.  It would be just as easy to pass them to the representation model and output model, if we find we need that in the future.

As an example of using this, I made the HDF5 dataset class support an optional `partial_charges` field.

We might consider getting rid of the hardcoded `q` and `s` arguments (I couldn't find any existing models that use them) and letting them get passed in `extra_args` if they're ever needed.

### DIFF
--- a/tests/test_prior_d2.py
+++ b/tests/test_prior_d2.py
@@ -80,6 +80,6 @@ def test_d2(test_case):
     batch = pt.tensor(data["batch"])
 
     y_init = pt.zeros_like(y_ref)
-    y_res = prior.post_reduce(y_init, z, pos, batch)
+    y_res = prior.post_reduce(y_init, z, pos, batch, {})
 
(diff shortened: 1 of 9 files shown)


raimis (approved):
Looks good to me

peastman:
Thanks!  We can remove them in a separate PR.

@PhilippThoelke does this look ok to you?

PhilippThoelke (approved):
Yes, looks good!

peastman:
Thanks!

PR #159, opened 2022-12-08
Created Coulomb prior

peastman:
This adds a Coulomb interaction.  At the moment it requires partial charges to be specified by the dataset.  In the future we'll want to let the model predict charges, but that will come later.

The interaction is applied to all pairs without any cutoff.  The point of this prior is to provide accurate long range interactions.  It does get scaled by `erf(alpha*r)`, which reduces the effect at short ranges and prevents it from diverging at r=0.  That also happens to be the scale factor of the reciprocal space term in Ewald summation.  Eventually we'll want to apply this to large systems that are partly modeled with ML and partly with a conventional force field.  The idea is that you'll be able to include the Coulomb prior simply by including the ML region in the reciprocal space PME calculation.

### DIFF
--- a/torchmdnet/priors/coulomb.py
+++ b/torchmdnet/priors/coulomb.py
@@ -44,4 +44,6 @@ class Coulomb(BasePrior):
         # appears twice.
         q = extra_args['partial_charges'][edge_index]
         energy = torch.erf(alpha*distance)*q[0]*q[1]/distance
-        return y + 0.5*(2.30707e-28/self.energy_scale/self.distance_scale)*scatter(energy, batch[edge_index[0]], dim=0, reduce="sum")
+        energy = 0.5*(2.30707e-28/self.energy_scale/self.distance_scale)*scatter(energy, batch[edge_index[0]], dim=0, reduce="sum")
+        energy = energy.reshape(y.shape)
+        return y + energy
--- a/torchmdnet/priors/coulomb.py
+++ b/torchmdnet/priors/coulomb.py
@@ -20,7 +20,7 @@ class Coulomb(BasePrior):
             distance_scale = dataset.distance_scale
         if energy_scale is None:
             energy_scale = dataset.energy_scale
-        self.distance = Distance(0, torch.inf, max_num_neighbors=100)
+        self.distance = Distance(0, torch.inf, max_num_neighbors=max_num_neighbors)
         self.alpha = alpha
         self.max_num_neighbors = max_num_neighbors
         self.distance_scale = float(distance_scale)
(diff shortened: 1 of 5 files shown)


peastman:
Does anyone have comments on this?

torchmdnet/priors/coulomb.py
```diff
@@ -0,0 +1,50 @@
...
+    distance_scale: multiply by this factor to convert coordinates stored in the dataset to meters
+    energy_scale: multiply by this factor to convert energies stored in the dataset to Joules (*not* J/mol)
+    """
+    def __init__(self, alpha, max_num_neighbors, distance_scale=None, energy_scale=None, dataset=None):
+        super(Coulomb, self).__init__()
+        if distance_scale is None:
+            distance_scale = dataset.distance_scale
+        if energy_scale is None:
+            energy_scale = dataset.energy_scale
+        self.distance = Distance(0, torch.inf, max_num_neighbors=100)
```
PhilippThoelke, line 23: Why are you hard coding `max_num_neighbors` instead of using the parameter?

peastman: Thanks for catching that.  That was just done for development.  I meant to create a parameter for it but forgot.

peastman: Fixed.

peastman:
Is it ok to merge this?

torchmdnet/priors/coulomb.py
```diff
@@ -0,0 +1,50 @@
...
+    distance_scale: multiply by this factor to convert coordinates stored in the dataset to meters
+    energy_scale: multiply by this factor to convert energies stored in the dataset to Joules (*not* J/mol)
+    """
+    def __init__(self, alpha, max_num_neighbors, distance_scale=None, energy_scale=None, dataset=None):
+        super(Coulomb, self).__init__()
+        if distance_scale is None:
+            distance_scale = dataset.distance_scale
+        if energy_scale is None:
+            energy_scale = dataset.energy_scale
+        self.distance = Distance(0, torch.inf, max_num_neighbors=max_num_neighbors)
```
RaulPPelaez, line 23: You can pass return_vecs=False since you are not using edge_vecs below.

torchmdnet/priors/coulomb.py
```diff
@@ -0,0 +1,50 @@
...
+from typing import Optional, Dict
+
+class Coulomb(BasePrior):
+    """This class implements a Coulomb potential, scaled by erf(alpha*r) to reduce its
+    effect at short distances.
+
+    To use this prior, the Dataset must include a field called `partial_charges` with each sample,
+    containing the partial charge for each atom.  It also must provide the following attributes.
+
+    distance_scale: multiply by this factor to convert coordinates stored in the dataset to meters
```
RaulPPelaez, line 14: Why not nanometers directly?

RaulPPelaez:
This LGTM, its merge from me.

torchmdnet/priors/coulomb.py
peastman: `return_vecs=False` is already the default.

peastman:
Thanks!

PR #161, opened 2023-01-24
Fix deprecated usage of np.int with numpy>1.23

RaulPPelaez:
Tests failed for me (with the suggested environment.yml, which installed numpy=1.24) due to usage of the [deprecated numpy.int](https://numpy.org/devdocs/release/1.20.0-notes.html#using-the-aliases-of-builtin-types-like-np-int-is-deprecated).
I replaced a couple of instances of np.int by just int, as suggested in the numpy docs.

All tests pass with the numpy 1.24.


stefdoerr (approved):
Looks good. Thanks!

PR #164, opened 2023-03-17
Update SPICE

raimis:
The SPICE dataset is now published on Zendo (https://zenodo.org/record/7606550)

- [x] Download the old versions from GitHub
- [x] Download the new versions from Zendo
- [x] Set default version to `1.1.3`


guillemsimeon:
It works now

PR #172, opened 2023-05-22
Update md17

AntonioMirarchi:
The file names used in the MD17 dataloader (torchmdnet/datasets/md17.py) have become obsolete. This pull request (PR) aims to update all the file names in accordance with the latest release available at http://www.quantum-machine.org/gdml/data/npz/.

### DIFF
--- a/torchmdnet/datasets/md17.py
+++ b/torchmdnet/datasets/md17.py
@@ -13,14 +13,14 @@ class MD17(InMemoryDataset):
     raw_url = "http://www.quantum-machine.org/gdml/data/npz/"
 
     molecule_files = dict(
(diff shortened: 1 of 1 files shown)


RaulPPelaez (approved):
Looks good to me, always have to change these -.-

AntonioMirarchi:
There is also the possibility to include more datasets like the revised ones.

RaulPPelaez:
Should I merge now or do you want to add those before?

AntonioMirarchi:
I looked closer at the other datasets, and to include them we need to change the dataloader because they are available on a different website and also the single arrays are named differently in the npz files respect to the original ones. So I think you can merge this for the moment

PR #174, opened 2023-05-23
Enable the Ace loader to return a conformation index

raimis:
Enable the Ace loader to return a conformation index

```python
ds = Ace('.', paths='dataset.h5')
for s in ds.sample_iter(mol_ids=True):
    print(s)

Data(y=[1, 1], pos=[12, 3], z=[12], neg_dy=[12, 3], q=0, pq=[12], dp=[3])
...

for s in ds.sample_iter(mol_ids=True):
    print(s)

Data(y=[1, 1], pos=[12, 3], z=[12], neg_dy=[12, 3], q=0, pq=[12], dp=[3], mol_id='C4H5N3', i_conf=10355)
...

```

### DIFF
--- a/torchmdnet/datasets/ace.py
+++ b/torchmdnet/datasets/ace.py
@@ -162,7 +162,7 @@ class Ace(Dataset):
                 mols = list(h5.values())[0].items()
                 load_confs = self._load_confs_2_0
             else:
-                raise RuntimeError(f"Unsuported layout verions: {version}")
+                raise RuntimeError(f"Unsupported layout version: {version}")
 
             # Iterate over the molecules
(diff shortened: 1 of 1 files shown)


RaulPPelaez (approved):
LGTM. 
There is no test coverage for this Dataset, right?

torchmdnet/datasets/ace.py
RaulPPelaez, line 1: For consistency you could change these lines:
https://github.com/torchmd/torchmd-net/pull/174/files#diff-56de3e28dd2b438171f4af7de59cc66305304d98e21b24faf4d744698d8c1aa3R269-R279
To use the i_conf property

PR #179, opened 2023-06-05
Fix neighbor backward

RaulPPelaez:
Trying to integrate OptimizedDistance into the equivariant transformer (ET) I realized my module did not supported double backwards (DB), which is required in ET when training with forces.

This PR adds support for double gradient to OptimizedDistance, and adds some tests for it using [torch.autograd.gradgradcheck](https://pytorch.org/docs/stable/generated/torch.autograd.gradgradcheck.html).

After some research I learned about [how to support DB in a custom extension](https://pytorch.org/tutorials/intermediate/custom_function_double_backward_tutorial.html#saving-intermediate-results). I rewrote the backwards pass in full pytorch to take advantage of autograd for the second gradient. Then I discovered Python [custom torch.Autograd.Function are not compatible with jit.script ](https://github.com/pytorch/pytorch/issues/22329), but it turns out[ it is compatible if you do it in C++](https://pytorch.org/tutorials/advanced/dispatcher.html#adding-autograd-support). So I translated the full pytorch backwards to C++. Quite the journey -.-


RaulPPelaez:
This is ready for review

torchmdnet/neighbors/neighbors_cuda.cu
```diff
@@ -2,45 +2,79 @@
...
+        Tensor neighbors, deltas, distances, i_curr_pair;
+        std::tie(neighbors, deltas, distances, i_curr_pair) =
+            call_forward_kernel(strategy, positions, batch, box_vectors, use_periodic, cutoff_lower,
+                                cutoff_upper, max_num_pairs, loop, include_transpose);
+        ctx->save_for_backward({neighbors, deltas, distances});
+        ctx->saved_data["num_atoms"] = positions.size(0);
+        return {neighbors, deltas, distances, i_curr_pair};
+    }
+
+    static tensor_list backward(AutogradContext* ctx, tensor_list grad_outputs) {
```
raimis, line 41: Is this compatible with CUDA Graphs?

torchmdnet/neighbors/neighbors_cuda_brute.cuh
```diff
@@ -48,67 +49,55 @@ __global__ void add_self_kernel(const int num_atoms, Accessor<scalar_t, 2> posit
...
+        PairListAccessor<scalar_t> list_accessor(list);
+        triclinic::Box<scalar_t> box(box_vectors, use_periodic);
+        const scalar_t cutoff_upper_ = cutoff_upper.to<scalar_t>();
+        const scalar_t cutoff_lower_ = cutoff_lower.to<scalar_t>();
+        TORCH_CHECK(cutoff_upper_ > 0, "Expected \"cutoff\" to be positive");
+        forward_kernel_brute<<<num_blocks, num_threads, 0, stream>>>(
+            num_all_pairs, get_accessor<scalar_t, 2>(positions), get_accessor<int64_t, 1>(batch),
+            cutoff_lower_ * cutoff_lower_, cutoff_upper_ * cutoff_upper_, list_accessor, box);
+        if (loop) {
+            const uint32_t num_threads = 256;
```
raimis, line 94: This is shadowing the previous definition.

torchmdnet/neighbors/neighbors_cpu.cpp
```diff
@@ -90,7 +90,9 @@ forward(const Tensor& positions, const Tensor& batch, const Tensor& box_vectors,
 }
 
 TORCH_LIBRARY_IMPL(torchmdnet_neighbors, CPU, m) {
-    m.impl("get_neighbor_pairs_brute", &forward);
-    m.impl("get_neighbor_pairs_shared", &forward);
-    m.impl("get_neighbor_pairs_cell", &forward);
+  m.impl("get_neighbor_pairs", [](std::string strategy,  const Tensor& positions, const Tensor& batch, const Tensor& box_vectors,
```
raimis, line 93: Maybe:
```c++
const std::string strategy
```

torchmdnet/neighbors/neighbors_cuda_brute.cuh
RaulPPelaez: This is by design, I noticed 256 works better for the self_kernel.

torchmdnet/neighbors/neighbors_cpu.cpp
RaulPPelaez: I was being evil and changing the value of strategy under some circumstances. I agree, lets do it the right way :P

torchmdnet/neighbors/neighbors_cuda_brute.cuh
peastman: It's still better not to shadow an earlier variable.  That tends to be confusing.

torchmdnet/neighbors/neighbors_cuda.cu
RaulPPelaez: It is not, sadly...
libtorch is lacking some convenient functions to facilitate that, like masked_index_add.
I am going to merge this and work on that in another PR.

raimis (approved):

PR #180, opened 2023-06-06
TN

guillemsimeon:
PR to discuss full integration of TN into TorchMD-NET


RaulPPelaez:
I believe tests fail because model.py expects the neighbor_embedding parameter to exist in the yaml, which your example yamls ommit

environment.yml
RaulPPelaez, line 1: Please use the latest environment. If you find it is not equivalent then that is a bug in TorchMD-Net.

torchmdnet/datasets/md17.py
RaulPPelaez, line 1: Why the large modifications to this file?

guillemsimeon: Old MD17 dataset class allowed training only on the original MD17 dataset. The new class (extracted from PyG MD17 dataset class) allows simultaneous training on original, revised and CCSD(T) dataset. TensorNet is benchmarked using the revised version.

environment.yml
guillemsimeon: Done

torchmdnet/datasets/md17.py
```diff
@@ -1,105 +1,151 @@
...
+        'aspirin': 'md17_aspirin.npz',
+        'salicylic acid': 'md17_salicylic.npz',
+        'malonaldehyde': 'md17_malonaldehyde.npz',
+        'ethanol': 'md17_ethanol.npz',
+        'toluene': 'md17_toluene.npz',
+        'paracetamol': 'paracetamol_dft.npz',
+        'azobenzene': 'azobenzene_dft.npz',
+        'revised benzene': 'rmd17_benzene.npz',
+        'revised uracil': 'rmd17_uracil.npz',
+        'revised naphthalene': 'rmd17_naphthalene.npz',
```
AntonioMirarchi, line 35: If I understood correctly, it's going to create a dir (in `self.root`) for each dataset using the key, so maybe use the underscore for the keys: 'revised_benzene', 'revised_uracil', etc.

torchmdnet/datasets/md17.py
```diff
@@ -1,105 +1,151 @@
...
-    def process(self):
-        for raw_path, processed_path in zip(self.raw_paths, self.processed_paths):
-            data_npz = np.load(raw_path)
-            z = torch.from_numpy(data_npz["z"]).long()
-            positions = torch.from_numpy(data_npz["R"]).float()
-            energies = torch.from_numpy(data_npz["E"]).float()
-            forces = torch.from_numpy(data_npz["F"]).float()
+    @property
+    def processed_dir(self) -> str:
+        return osp.join(self.root, self.name, 'processed')
```
AntonioMirarchi, line 93: Now using this, you put your processed data in: self.root/self.name/processed. 
I don't know if this could help: `return osp.join(self.root, 'processed', self.name)` it's going to create a processed dir for your specific data in data/processed so you can keep the conventional data gerarchy: self.root/raw and self.root/processed.

torchmdnet/scripts/train.py
```diff
@@ -81,6 +82,9 @@ def get_args():
     parser.add_argument('--distance-influence', type=str, default='both', choices=['keys', 'values', 'both', 'none'], help='Where distance information is included inside the attention')
     parser.add_argument('--attn-activation', default='silu', choices=list(act_class_mapping.keys()), help='Attention activation function')
     parser.add_argument('--num-heads', type=int, default=8, help='Number of attention heads')
+    
+    # TensorNet specific
+    parser.add_argument('--equivariance-invariance_group', type=str, default='O(3)', help='Equivariance and invariance group of TensorNet')
```
peastman, line 87: I think that should be `'--equivariance-invariance-group'`?  With a hyphen rather than an underscore.

peastman:
This model isn't compatible with TorchScript.  When I try to compile it with `torch.jit.script()` it gets an exception.

```
RuntimeError: 
cannot statically infer the expected size of a list in this context:
  File "/home/peastman/workspace/torchmd-net/torchmdnet/models/tensornet.py", line 16
def vector_to_skewtensor(vector):
    tensor = torch.cross(*torch.broadcast_tensors(vector[...,None], torch.eye(3,3, device=vector.device)[None,None]))
                         ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
    return tensor.squeeze(0)
```

RaulPPelaez:
Add tensornet here so the tests pick it up:
https://github.com/torchmd/torchmd-net/blob/e20876ff79fe952db25e5fc55cc051a919ef218f/torchmdnet/models/__init__.py#L1

guillemsimeon:
@peastman the model is indeed not compatible with TorchScript. @RaulPPelaez modified the model at some point such that it was TorchScript compatible. However, we didn't have time to retest performances on datasets, but we observed a 10-15% speed-up. @RaulPPelaez will be working on that.

RaulPPelaez:
Test are failing because of the equivariance/invariance arg https://github.com/torchmd/torchmd-net/actions/runs/5196861671/jobs/9371007295?pr=180#step:7:4296

guillemsimeon:
@RaulPPelaez I think it is due to the fact that load_example_args uses an ET example yaml, which does not contain equivariance_invariance_group arg.

guillemsimeon:
@RaulPPelaez A possible solution I think it would be to create an example yaml with all arguments in it, independent of the architecture. For models not using specific arguments (such as attention heads and so on for graphnet and TN) these would not be considered.

RaulPPelaez:
I believe it would be best to use a valid TN yaml in that case, load_example_args can be modified accordingly:
https://github.com/torchmd/torchmd-net/blob/237b4fe1bf0ce52721b42fe51f35efca018753f9/tests/utils.py#L7-L19
I made it so it loads a tn file for tensornet tests.
BTW, for consistency maybe tensornet example yamls should be called "TN_*yaml"

RaulPPelaez:
I turned off the test that checks jit script compatibility for TN. Eventually I will replace the offending functions in the architecture, I want to make sure I do not change accuracy.

RaulPPelaez:
Next thing is this test_forward_output
https://github.com/torchmd/torchmd-net/blob/237b4fe1bf0ce52721b42fe51f35efca018753f9/tests/test_model.py#L62
If overwrite_reference is True then the test will overwrite the expected.pkl file with the result of forwarding some random numbers:
https://github.com/torchmd/torchmd-net/blob/237b4fe1bf0ce52721b42fe51f35efca018753f9/tests/test_model.py#L85-L90
You run this test by going into the test folder and running:
```shell
$ pytest -k forward_output test_model.py
```
You must go to the torchmd-net installation you used for reporting results and:
1. Set overwrite_reference=True in test_model.py
2. Run the test
3. Add the generated expected.pkl file to this PR

By doing this, you get some assurance that any changes you make to the arch or the environment do not break your model. Later on, the gradient tests will check that backwards is consistent too.

guillemsimeon:
I would deal with JIT compatibility in a separate PR or something like that. Checks will need to be made for accuracies and so on.

RaulPPelaez:
Fine for me, I am working on it here: https://github.com/RaulPPelaez/torchmd-net/tree/tn_script
Will PR it after this is merged.

torchmdnet/models/tensornet.py
```diff
@@ -0,0 +1,397 @@
...
+        max_num_neighbors=64,
+        return_vecs=True,
+        loop=True,
+        trainable_rbf=False,
+        max_z = 128,
+        equivariance_invariance_group = "O(3)",
+    ):
+        super(TensorNet, self).__init__()
+        
+        assert rbf_type in rbf_class_mapping, (
```
RaulPPelaez, line 64: Lets add some documentation similar to ET:
https://github.com/torchmd/torchmd-net/blob/237b4fe1bf0ce52721b42fe51f35efca018753f9/torchmdnet/models/torchmd_et.py#L16-L53
In this case a reference to the paper will also be cool.

torchmdnet/models/tensornet.py
```diff
@@ -55,20 +55,52 @@ def tensor_norm(tensor):
 
 
 class TensorNet(nn.Module):
+    r"""TensorNet's architecture.
```
RaulPPelaez, line 58: Reference the article somehow. A link to a doi at least.

torchmdnet/models/tensornet.py
```diff
@@ -55,20 +55,52 @@ def tensor_norm(tensor):
...
+            (default: :obj:`5.0`)
+        max_z (int, optional): Maximum atomic number. Used for initializing embeddings.
+            (default: :obj:`100`)
+        max_num_neighbors (int, optional): Maximum number of neighbors to return for a
+            given node/atom when constructing the molecular graph during forward passes.
+            This attribute is passed to the torch_cluster radius_graph routine keyword
+            max_num_neighbors, which normally defaults to 32. Users should set this to
+            higher values if they are using higher upper distance cutoffs and expect more
+            than 32 neighbors per node/atom.
+            (default: :obj:`32`)
```
RaulPPelaez, line 85: Its actually 64 now.

torchmdnet/models/tensornet.py
```diff
@@ -55,20 +55,52 @@ def tensor_norm(tensor):
...
+        trainable_rbf (bool, optional): Whether to train RBF parameters with
+            backpropagation. (default: :obj:`True`)
+        activation (string, optional): The type of activation function to use.
+            (default: :obj:`"silu"`)
+        cutoff_lower (float, optional): Lower cutoff distance for interatomic interactions.
+            (default: :obj:`0.0`)
+        cutoff_upper (float, optional): Upper cutoff distance for interatomic interactions.
+            (default: :obj:`5.0`)
+        max_z (int, optional): Maximum atomic number. Used for initializing embeddings.
+            (default: :obj:`100`)
```
RaulPPelaez, line 78: Its 128 now.

torchmdnet/models/tensornet.py
```diff
@@ -55,20 +55,52 @@ def tensor_norm(tensor):
...
+        rbf_type (string, optional): The type of radial basis function to use.
+            (default: :obj:`"expnorm"`)
+        trainable_rbf (bool, optional): Whether to train RBF parameters with
+            backpropagation. (default: :obj:`True`)
+        activation (string, optional): The type of activation function to use.
+            (default: :obj:`"silu"`)
+        cutoff_lower (float, optional): Lower cutoff distance for interatomic interactions.
+            (default: :obj:`0.0`)
+        cutoff_upper (float, optional): Upper cutoff distance for interatomic interactions.
+            (default: :obj:`5.0`)
```
RaulPPelaez, line 76: Its 4.5

torchmdnet/models/tensornet.py
```diff
@@ -55,20 +55,52 @@ def tensor_norm(tensor):
...
+        hidden_channels (int, optional): Hidden embedding size.
+            (default: :obj:`128`)
+        num_layers (int, optional): The number of interaction layers.
+            (default: :obj:`6`)
+        num_rbf (int, optional): The number of radial basis functions :math:`\mu`.
+            (default: :obj:`50`)
+        rbf_type (string, optional): The type of radial basis function to use.
+            (default: :obj:`"expnorm"`)
+        trainable_rbf (bool, optional): Whether to train RBF parameters with
+            backpropagation. (default: :obj:`True`)
```
RaulPPelaez, line 70: Its False.

RaulPPelaez (approved):

RaulPPelaez:
All good! lets merge.
Thanks @guillemsimeon!

PR #182, opened 2023-06-06
Support for float64

RaulPPelaez:
I added a new parameter, dtype, to control whether TorchMD_Net uses float or double.
The value can be passed along the rest to create_model, but it defaults to float if not present. Can be either a string or a torch.dtype.

I also added some comments here and there.

To showcase I added a test that checks the correctness of gradients for all models using torch.autograd.gradcheck.


torchmdnet/models/model.py
```diff
@@ -7,11 +7,27 @@
...
+    ----------
+        args (dict): Arguments for the model.
+        prior_model (nn.Module, optional): Prior model to use. Defaults to None.
+        mean (torch.Tensor, optional): Mean of the training data. Defaults to None.
+        std (torch.Tensor, optional): Standard deviation of the training data. Defaults to None.
+    Returns
+    -------
+        nn.Module: An instance of the TorchMD_Net model.
+    """
+    args["dtype"] = "float" if "dtype" not in args else args["dtype"]
```
raimis, line 28: It should be `float32` to avoid ambiguity.

torchmdnet/scripts/train.py
```diff
@@ -66,6 +66,7 @@ def get_args():
     parser.add_argument('--prior-model', type=str, default=None, choices=priors.__all__, help='Which prior model to use')
 
     # architectural args
+    parser.add_argument('--dtype', type=str, default="float", choices=list(dtype_mapping.keys()), help='Floating point precision. Can be float or float64')
```
raimis, line 69: The same, `float32`

peastman:
This tensor should also get a dtype:

https://github.com/torchmd/torchmd-net/blob/e20876ff79fe952db25e5fc55cc051a919ef218f/torchmdnet/priors/d2.py#L49

PR #185, opened 2023-06-08
fix torch tensor warning in spice dataloader

AntonioMirarchi:
To avoid the warning in PyTorch regarding tensor creation from a list of numpy (usually regards only `mol[conformations]`), it is recommended to create a torch tensor from a numpy array.


RaulPPelaez (approved):
LGTM. One less warning to go. Thanks!

PR #190, opened 2023-06-19
Wandb resume

AntonioMirarchi:
In certain scenarios, there is a need to resume training without creating a new wandb logger, for example killed or crashed process. The purpose of this pull request is to incorporate this functionality directly into the tochmd-net code, enabling users to have this possibility directly from the YAML conf file. To achieve this, two new arguments have been introduced: 

- `wandb-resume` (boolean); 
- `wandb-id`, the ID of the run to be resumed (it's possible to retrieve it from the project table on the wandb-page);

### DIFF
--- a/torchmdnet/scripts/train.py
+++ b/torchmdnet/scripts/train.py
@@ -98,6 +98,8 @@ def get_args():
     parser.add_argument('--wandb-use', default=False, type=bool, help='Defines if wandb is used or not')
     parser.add_argument('--wandb-name', default='training', type=str, help='Give a name to your wandb run')
     parser.add_argument('--wandb-project', default='training_', type=str, help='Define what wandb Project to log to')
+    parser.add_argument('--wandb-resume', default=False, type=bool, help='Defines if wandb run should be resumed')
+    parser.add_argument('--wandb-id', default=None, type=str, help='Define what wandb run to resume, you can find the id from the wandb table')
     parser.add_argument('--tensorboard-use', default=False, type=bool, help='Defines if tensor board is used or not')
 
     # fmt: on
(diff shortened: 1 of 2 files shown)


torchmdnet/scripts/train.py
```diff
@@ -98,6 +98,8 @@ def get_args():
     parser.add_argument('--wandb-use', default=False, type=bool, help='Defines if wandb is used or not')
     parser.add_argument('--wandb-name', default='training', type=str, help='Give a name to your wandb run')
     parser.add_argument('--wandb-project', default='training_', type=str, help='Define what wandb Project to log to')
+    parser.add_argument('--wandb-resume', default=False, type=bool, help='Defines if wandb run should be resumed')
+    parser.add_argument('--wandb-id', default=None, type=str, help='Define what wandb run to resume, you can find the id from the wandb table')
```
RaulPPelaez, line 102: You could have just one argument: --wandb-resume-from-id. I see no possibility for wandb-resume to be present when wandb-id is not.

torchmdnet/scripts/train.py
```diff
@@ -98,6 +98,7 @@ def get_args():
     parser.add_argument('--wandb-use', default=False, type=bool, help='Defines if wandb is used or not')
     parser.add_argument('--wandb-name', default='training', type=str, help='Give a name to your wandb run')
     parser.add_argument('--wandb-project', default='training_', type=str, help='Define what wandb Project to log to')
+    parser.add_argument('--wandb-resume-from-id', default=None, type=str, help='Resume a wandb run from a given run id, retrieve the run id from the wandb dashboard')
```
RaulPPelaez, line 101: Resume a wandb run from a given run id. The id can be retrieved from the wandb dashboard

RaulPPelaez (approved):
Looks good, shall I merge?

PR #191, opened 2023-06-20
[DNM] Make setup.py compile/install  Cpp extensions

RaulPPelaez:
Currently cpp extensions are JIT compiled when first imported. This causes a lot of time consuming recompilation.
I am trying to instead compile them to a library in setup.py, but failing miserably. 

I followed this pytorch tutorial: https://pytorch.org/tutorials/advanced/cpp_extension.html

My modifications correctly generate a library in the right place, but when I try to import it I get an error:
```
>   from torchmdnet_neighbors import get_neighbor_pairs
E   ImportError: dynamic module does not define module export function (PyInit_torchmdnet_neighbors)
```
Dissasembling this library shows that there is in fact no PyInit function. I do not know what I am missing here. 

According to the tutorial, torch must be imported first, which I am doing to no avail.

I am PRing here to see if anyone has any idea!

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,5 +1,8 @@
 import subprocess
 from setuptools import setup, find_packages
+import os
+import torch
+from torch.utils.cpp_extension import CppExtension, BuildExtension, CUDAExtension
 
 try:
     version = (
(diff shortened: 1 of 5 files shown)


RaulPPelaez:
Closing as this is superseded by #216

Issue #168, opened 2023-04-25
Updating torch lightning


RaulPPelaez:
In a quest to use the latest versions of the different dependencies (mainly for torch2) I have encountered an issue with torch-lightning.
The API has changed in some slight ways, in particular this reset line here has been deprecated:
https://github.com/torchmd/torchmd-net/blob/e02c076aefea88ed39a5b161ba48113b31ba331e/torchmdnet/module.py#L160-L163
There is a succint message refering to this in the docs:
```markdown
   * - used ``trainer.reset_*_dataloader()`` methods
     - use  ``Loop.setup_data()`` for the top-level loops
     - `PR16726`_
```
Pointing to this PR: https://github.com/Lightning-AI/lightning/pull/16726

I do not know, however, what this line is doing exactly and I do not understand the comment.
Do you have some pointers?
cc @PhilippThoelke

PhilippThoelke:
To evaluate the test set every couple of epochs during training we add it as a val_dataloader. In the test epochs we then have two dataloaders (validation + test) instead of only the validation dataloader. The deprecated function resets the Trainer's internal state of which data to use for what, allowing us to evaluate on the test set only every couple of epochs but evaluate on the validation set every epoch. You'd have to make sure that the suggested replacement method achieves the same thing without breaking or slowing down anything else.

PR #195, opened 2023-06-26
Make TensorNet CUDA-graph compatible

RaulPPelaez:
This PR is an effort to optimize TensorNet. Note that I merged here also all the changes in #186. 

I have focused on inference of small molecules (such as Alanine, 22 atoms), in which we saw the CUDA kernel launching overhead was tremendous:
image
My strategy was to simply modify what was needed for ```torch.compile(backend="inductor", mode="reduce-overhead")``` to put as much as possible into a CUDA graph.
 
I replaced some blocking operations and I was able to convince TorchDynamo to put the bulk of the forward pass into a CUDA graph (orange is a graph):
image
 Zooming in, I am having trouble with making TorchDynamo understand the neighbor kernels:
image
Alas, TorchDynamo refuses to graph the backwards pass. But I know it is possible to graph it, since I can do so manually.

Anyhow, I believe this is a problem that will solve itself in time as TorchDynamo improves its support for things like Dynamic shapes and extensions. And even today, the speedup is impressive.

I am not sure how to apply optimizations such as CUDA graphs for training, where shapes are dynamic. However, in those situations the changes sin this PR will just not improve things as much. TorchDynamo refuses to even process the model when used during training, complaining about dynamic shapes.

You will notice the benchmark tries to compile the model by doing this:
```python
model = torch.compile(model, backend="inductor", mode="max-autotune")
```
This line fails miserably when using the current environment, I had to use pytorch-nightly (which provides version 2.1.0):
```shell
conda install pytorch torchvision torchaudio pytorch-cuda=12.1 -c pytorch-nightly -c nvidia
```  
Which made me install the rest of the torch friends with pip:
```shell
$ pip list | grep torch
pytorch-lightning       1.6.3
torch                   2.1.0.dev20230619
torch-cluster           1.6.1
torch-geometric         2.3.1
torch-scatter           2.1.1
torchmdnet              0.2.5
torchmetrics            0.11.4
```
Luckily these compile just fine with pip.
With a little luck after this https://github.com/conda-forge/pytorch-cpu-feedstock/pull/172 is merged we can update the env and stay in conda-forge.


peastman:
What happens if you compile to torchscript and don't try to use graphs?  Does that help training speed?

RaulPPelaez:
I was not able to train on a TorchScripted model. I suspect there might be some issue with the old-ish torch lightning version currently used. However upgrading it is dauting #168.
When trying to train with an scripted model (this is not particular to TensorNet) I get this error:
https://github.com/torchmd/torchmd-net/pull/186#issuecomment-1587513829
I have no idea where it comes from...

guillemsimeon (approved):
Thanks a lot, Raul

guillemsimeon:
Is it necessary to do something specific to run inference with all your optimizations? Or they will work out-of-the-box?

torchmdnet/models/tensornet.py
```diff
@@ -188,19 +205,28 @@ def forward(
...
+            # WARNING: This can hurt performance if max_num_pairs >> actual_num_pairs
+            edge_index = edge_index * mask
+            edge_weight = edge_weight * mask[0]
+            edge_vec = edge_vec * mask[0].unsqueeze(-1).expand_as(edge_vec)
         edge_attr = self.distance_expansion(edge_weight)
-        # Embedding from edge-wise tensors to node-wise tensors
+        mask = edge_index[0] != edge_index[1]
+        # Normalizing edge vectors by their length can result in NaNs, breaking Autograd.
+        # I avoid dividing by zero by setting the weight of self edges and self loops to 1
+        edge_vec = edge_vec / torch.masked_scatter(torch.ones_like(edge_weight),
```
raimis, line 226: I think `maked_scatter` is working differently than you think (https://pytorch.org/docs/stable/generated/torch.Tensor.masked_scatter.html)

This should be a correct implementation
```
diff --git a/torchmdnet/models/tensornet.py b/torchmdnet/models/tensornet.py
index 045219b..6b005b1 100644
--- a/torchmdnet/models/tensornet.py
+++ b/torchmdnet/models/tensornet.py
@@ -220,12 +220,10 @@ class TensorNet(nn.Module):
             edge_weight = edge_weight * mask[0]
             edge_vec = edge_vec * mask[0].unsqueeze(-1).expand_as(edge_vec)
         edge_attr = self.distance_expansion(edge_weight)
-        mask = edge_index[0] != edge_index[1]
         # Normalizing edge vectors by their length can result in NaNs, breaking Autograd.
         # I avoid dividing by zero by setting the weight of self edges and self loops to 1
-        edge_vec = edge_vec / torch.masked_scatter(torch.ones_like(edge_weight),
-            mask, edge_weight
-        ).unsqueeze(1)
+        mask = edge_index[0] == edge_index[1]
+        edge_vec = edge_vec / edge_weight.masked_fill(mask, 1).unsqueeze(1)
         X = self.tensor_embedding(z, edge_index, edge_weight, edge_vec, edge_attr)
         for layer in self.layers:
             X = layer(X, edge_index, edge_weight, edge_attr)

```

RaulPPelaez: The documentation is actually super misleading I believe https://github.com/pytorch/pytorch/issues/82377

RaulPPelaez: But looking at it your implementation looks better and masked_scatter is problematic with torch.compile. Lets use yours. Thanks!

raimis:
To make the model fully compatible with CUDA Graph, I need replace `scatter` with reduction:
```diff
diff --git a/torchmdnet/models/output_modules.py b/torchmdnet/models/output_modules.py
index d3d7e1b..29edf12 100644
--- a/torchmdnet/models/output_modules.py
+++ b/torchmdnet/models/output_modules.py
@@ -25,7 +25,8 @@ class OutputModel(nn.Module, metaclass=ABCMeta):
         return
 
     def reduce(self, x, batch):
-        return scatter(x, batch, dim=0, reduce=self.reduce_op)
+        # return scatter(x, batch, dim=0, reduce=self.reduce_op)
+        return torch.sum(x)
 
     def post_reduce(self, x):
         return x
```
Of course, this is assumes that there is one molecule and the reduction is summing.

RaulPPelaez:
You can give scatter the argument dim_size with the number of different batches and this will be CUDA-graph compatible.
By default, scatter computes dim_size=batch.max()+1, which causes a sync. 
In the current code, just give it a 1 if you only have one molecule.

In general, not having access to the number of batches is a pain for CUDA graphs.
If there is only one then the batch argument can be None, which can be detected and leveraged without syncs. Otherwise unless we add an argument "num_molecules" I see no way out of a sync.

raimis:
OK! This is enough to fix graphing, but we need a general solution:
```diff
diff --git a/torchmdnet/models/output_modules.py b/torchmdnet/models/output_modules.py
index d3d7e1b..45e66eb 100644
--- a/torchmdnet/models/output_modules.py
+++ b/torchmdnet/models/output_modules.py
@@ -25,7 +25,7 @@ class OutputModel(nn.Module, metaclass=ABCMeta):
         return
 
     def reduce(self, x, batch):
-        return scatter(x, batch, dim=0, reduce=self.reduce_op)
+        return scatter(x, batch, dim=0, dim_size=1, reduce=self.reduce_op)
 
     def post_reduce(self, x):
         return x
```

RaulPPelaez:
Great! Did you found out anything about the ungraphable operation in backpropagation?

raimis:
With https://github.com/torchmd/torchmd-net/pull/195#discussion_r1325742792 and https://github.com/torchmd/torchmd-net/pull/195#issuecomment-1719195035, I can graph forward and backward passes, but the let me check in the full pipeline.

RaulPPelaez:
Awesome! masked_scatter has been giving me a hard time from the start, I can believe that being the offending function. Thanks for the replacement!

raimis (approved):

raimis:
The CUDA Graphs work, except `OutputModel`, which will be addressed in #214 or other.

PR #200, opened 2023-07-03
Make all tensor representations 4x4 matrices

guillemsimeon:
This PR is a slight modification of TensorNet, which should make current 3x3 tensors become 4x4 (the original 3x3 padded with zeros to become 4x4), such that mathematically operations are equivalent on the 3x3 blocks but we enable the use of tensor cores.

@RaulPPelaez , there is some dtype that should be taken care of.

### DIFF
--- a/torchmdnet/models/tensornet.py
+++ b/torchmdnet/models/tensornet.py
@@ -20,7 +20,7 @@ def vector_to_skewtensor(vector):
             zero,
             -vector[:, 2],
             vector[:, 1],
-            zero
+            zero,
             vector[:, 2],
             zero,
             -vector[:, 0],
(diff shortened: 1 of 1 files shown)


guillemsimeon:
I expect the equivariance test on TensorNet to not work because of new shapes.

guillemsimeon:
Also, there are several ways to create this 4x4 representations by padding zeros. Which is the computationally most efficient way I don't know.

RaulPPelaez:
@guillemsimeon, I benchmarked this some time ago and I believe it is really not the way to go about this since it just eats more memory and it is just slower. I think we should close this, unless you want to keep it around as a playground for some other ideas.

PR #204, opened 2023-07-18
Fix neighbor extension compilation looking for non-existent backwards.cu

RaulPPelaez:
The neighbor extension compilation looks for a file that does not exist anymore.


guillemsimeon (approved):
;)

PR #206, opened 2023-07-19
Implement element filtering in the Ace datasets

raimis:
Example: load sample containing only H and C
```python
from torchmdnet.datasets import Ace

ds = Ace(".", paths="dataset.h5", atomic_numbers=[1, 6])
```

### DIFF
--- a/torchmdnet/datasets/ace.py
+++ b/torchmdnet/datasets/ace.py
@@ -15,15 +15,17 @@ class Ace(Dataset):
         pre_transform=None,
         pre_filter=None,
         paths=None,
+        atomic_numbers=None,
         max_gradient=None,
         subsample_molecules=1,
     ):
         assert isinstance(paths, (str, list))
 
-        arg_hash = f"{paths}{max_gradient}{subsample_molecules}"
+        arg_hash = f"{paths}{atomic_numbers}{max_gradient}{subsample_molecules}"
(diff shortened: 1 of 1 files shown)


torchmdnet/datasets/ace.py
```diff
@@ -15,15 +15,17 @@ def __init__(
...
         pre_filter=None,
         paths=None,
+        atomic_numbers=None,
         max_gradient=None,
         subsample_molecules=1,
     ):
         assert isinstance(paths, (str, list))
 
-        arg_hash = f"{paths}{max_gradient}{subsample_molecules}"
+        arg_hash = f"{paths}{atomic_numbers}{max_gradient}{subsample_molecules}"
```
RaulPPelaez, line 24: atomic_numbers is a list, this string will be something like:
```"some_path[1, 2, 3]None1"```
Previously it had no spaces, now it does. Is this ok?
Also, atomic_numbers can be [1,2] in some instance and [2,1] in another.
While being functionally equivalent they will have different hashes. Is this what you want?
Otherwise a simple sort fixes it.

raimis: Store the elements as a set. Now the hash will be the same for different atom order.

raimis:
@RaulPPelaez at the end, I don't need this.

PR #208, opened 2023-08-04
Remove dtype parameter, use previously existing "precision" instead

RaulPPelaez:
The dtype parameter was causing some issues (see #205).
I realized that there is already a "pecision" parameter that can be used for the same thing.
This PR removes the dtype argument and uses precision instead, which can be 16, 32 or 64.

@AntonioMirarchi could you check that the issues you were seeing are gone with this PR?

### DIFF
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -25,7 +25,6 @@ def create_model(args, prior_model=None, mean=None, std=None):
     -------
         nn.Module: An instance of the TorchMD_Net model.
     """
-    args["precision"] = 32 if "precision" not in args else args["precision"]
     dtype = dtype_mapping[args["precision"]]
     shared_args = dict(
         hidden_channels=args["embedding_dimension"],
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -25,8 +25,8 @@ def create_model(args, prior_model=None, mean=None, std=None):
     -------
         nn.Module: An instance of the TorchMD_Net model.
     """
-    args["dtype"] = "float32" if "dtype" not in args else args["dtype"]
-    args["dtype"] = dtype_mapping[args["dtype"]] if isinstance(args["dtype"], str) else args["dtype"]
+    args["precision"] = 32 if "precision" not in args else args["precision"]
+    args["dtype"] = dtype_mapping[args["precision"]]
     shared_args = dict(
         hidden_channels=args["embedding_dimension"],
         num_layers=args["num_layers"],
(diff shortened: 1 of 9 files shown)


AntonioMirarchi:
Yes, let me to run a training and I will let you know. Looks good to me!

torchmdnet/models/model.py
```diff
@@ -25,8 +25,8 @@ def create_model(args, prior_model=None, mean=None, std=None):
     -------
         nn.Module: An instance of the TorchMD_Net model.
     """
-    args["dtype"] = "float32" if "dtype" not in args else args["dtype"]
-    args["dtype"] = dtype_mapping[args["dtype"]] if isinstance(args["dtype"], str) else args["dtype"]
+    args["precision"] = 32 if "precision" not in args else args["precision"]
+    args["dtype"] = dtype_mapping[args["precision"]]
```
AntonioMirarchi, line 29: Do you still need this line? You should remove the `args["dtype"]` since you already removed it from the argparser in train.py

RaulPPelaez: There needs to be still a way to send dtype down the line. Maybe I get away with just creating a local variable "dtype" in the constructor and giving that to the different modules.
Do you think this current way is problematic?

RaulPPelaez:
I worked with @AntonioMirarchi here to include the possibility of training in double precision.
Lightning complains when the module is set to double but the DataModule provides single precision inputs.

Ideally every dataset class should process their corresponding files in either a user-provided dtype or just float64 and have the "get" method cast to whatever type is needed.
However, this is a huge undertaking due to the large amount of datasets currently available (which most just read/write files in float32 and all of them provide float32 in their get method). 

Instead I opted for writing a dataset wrapper that is used by the DataModule:
```python
class FloatCastDatasetWrapper(Dataset):
    def __init__(self, dataset, dtype=torch.float64):
        super(FloatCastDatasetWrapper, self).__init__(dataset.root, dataset.transform, dataset.pre_transform, dataset.pre_filter)
        self.dataset = dataset
        self.dtype = dtype

    def len(self):
        return len(self.dataset)

    def get(self, idx):
        data = self.dataset.get(idx)
        for key, value in data:
            if torch.is_tensor(value) and torch.is_floating_point(value):
                setattr(data, key, value.to(self.dtype))
        return data
    def __getattr__(self, name):
        # Check if the attribute exists in the underlying dataset
        if hasattr(self.dataset, name):
            return getattr(self.dataset, name)
        raise AttributeError(f"'{type(self).__name__}' and its underlying dataset have no attribute '{name}'")
```
This simply intercepts the get method and casts every tensor in the data to the correct type.

Bonus points: Should the need arise it would be easy now to enable training/inference in other floating types, such as bfloat16 or even lower precision stuff like nf4

AntonioMirarchi (approved):
LGTM! I trained without any problem and error using single and double precision.

RaulPPelaez:
@guillemsimeon @raimis please review

guillemsimeon (approved):

PR #210, opened 2023-08-08
Updating to Lightning 2.0

RaulPPelaez:
This is an effort to update the pytorch lightning dependency (see #168 )
A lot has changed since the LNNP module was written (current version is 1.6.4) and I expect this update would increase compatibility with some functionalities we seek (AKA torch compile/torchscript training).

### DIFF
--- a/torchmdnet/data.py
+++ b/torchmdnet/data.py
@@ -92,10 +92,10 @@ class DataModule(LightningDataModule):
 
     def val_dataloader(self):
         loaders = [self._get_dataloader(self.val_dataset, "val")]
-        if (
-            len(self.test_dataset) > 0
-            and (self.trainer.current_epoch + 1) % self.hparams["test_interval"] == 0
-        ):
+        # To allow to report the performance on the testing dataset during training
+        # we send the trainer two dataloaders every few steps and modify the
+        # validation step to understand the second dataloader as test data.
+        if self._is_test_during_training_epoch():
             loaders.append(self._get_dataloader(self.test_dataset, "test"))
         return loaders
 
@@ -116,13 +116,16 @@ class DataModule(LightningDataModule):
     def std(self):
         return self._std
 
-    def _get_dataloader(self, dataset, stage, store_dataloader=True):
-        store_dataloader = (
-            store_dataloader and self.trainer.reload_dataloaders_every_n_epochs <= 0
+    def _is_test_during_training_epoch(self):
+        return (
+            len(self.test_dataset) > 0
+            and self.hparams["test_interval"] > 0
+            and self.trainer.current_epoch > 0
+            and self.trainer.current_epoch % self.hparams["test_interval"] == 0
         )
+
+    def _get_dataloader(self, dataset, stage, store_dataloader=True):
         if stage in self._saved_dataloaders and store_dataloader:
-            # storing the dataloaders like this breaks calls to trainer.reload_train_val_dataloaders
-            # but makes it possible that the dataloaders are not recreated on every testing epoch
             return self._saved_dataloaders[stage]
 
         if stage == "train":
--- a/torchmdnet/data.py
+++ b/torchmdnet/data.py
@@ -3,8 +3,8 @@ from tqdm import tqdm
 import torch
 from torch.utils.data import Subset
 from torch_geometric.loader import DataLoader
-from pytorch_lightning import LightningDataModule
-from pytorch_lightning.utilities import rank_zero_warn
+from lightning import LightningDataModule
+from lightning_utilities.core.rank_zero import rank_zero_warn
 from torchmdnet import datasets
 from torch_geometric.data import Dataset
 from torchmdnet.utils import make_splits, MissingEnergyException
--- a/torchmdnet/module.py
+++ b/torchmdnet/module.py
@@ -101,7 +101,9 @@ class LNNP(LightningModule):
         loss_name = loss_fn.__name__
         if self.hparams.derivative and "neg_dy" in batch:
             loss_neg_y = loss_fn(neg_y, batch.neg_dy)
-            loss_neg_y = self._update_loss_with_ema(stage, "neg_dy", loss_name, loss_neg_y)
+            loss_neg_y = self._update_loss_with_ema(
+                stage, "neg_dy", loss_name, loss_neg_y
+            )
         if "y" in batch:
             loss_y = loss_fn(y, batch.y)
             loss_y = self._update_loss_with_ema(stage, "y", loss_name, loss_y)
@@ -115,8 +117,12 @@ class LNNP(LightningModule):
         #   loss_name: name of the loss function
         #   loss: loss value
         alpha = getattr(self.hparams, f"ema_alpha_{type}")
-        if stage in ["train", "val"] and  alpha < 1:
-            ema = self.ema[stage][type][loss_name] if loss_name in self.ema[stage][type] else loss.detach()
+        if stage in ["train", "val"] and alpha < 1:
+            ema = (
+                self.ema[stage][type][loss_name]
+                if loss_name in self.ema[stage][type]
+                else loss.detach()
+            )
             loss = alpha * loss + (1 - alpha) * ema
             self.ema[stage][type][loss_name] = loss.detach()
         return loss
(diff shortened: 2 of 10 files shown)


RaulPPelaez:
The resume_from_checkpoint option in train.py has disappeared:
```python

    trainer = pl.Trainer(
        strategy=DDPStrategy(find_unused_parameters=False),
        max_epochs=args.num_epochs,
        accelerator="auto",
        devices=args.ngpus,
        num_nodes=args.num_nodes,
        default_root_dir=args.log_dir,
        #resume_from_checkpoint=None if args.reset_trainer else args.load_model,
        callbacks=[early_stopping, checkpoint_callback],
        logger=_logger,
        precision=args.precision,
        gradient_clip_val=args.gradient_clipping,
        inference_mode=False
    )
```
Trying to see how one is supposed to replace it

AntonioMirarchi:
From LightningDeprecationWarning: Setting `Trainer(resume_from_checkpoint=)` is deprecated in v1.5 and will be removed in v1.7. Please pass `Trainer.fit(ckpt_path=)` directly instead.
So it could be something like this: 
```
if not args.load_model:
   trainer.fit(model, data)
else:
   trainer.fit(model, data, ckpt_path=load_model)
```

RaulPPelaez:
I found this issue in which they discuss why Lighting does not want to support testing during the training loop:
https://github.com/Lightning-AI/lightning/issues/9254
They also describe the trick that is being used now of adding the test dataloader as a validation dataloader so you can use it during training.

However, I do not understand why we are paying such a high price in performance to test during training. @PhilippThoelke , you implemented this AFAIK, would you provide some insights?

PhilippThoelke:
Testing during training was useful since it can be difficult to estimate model performance on val loss, when val loss is also used to adjust the learning rate (e.g. through the ReduceLROnPlateau schedule). This is particularly important for fast prototyping and architecture development. I wasn't aware that this slows down training by 6x, that is insane! How much of that is actually due to the testing though, and how much are improvements in Lightning 2.0?
I don't know if that lr scheduler is still relevant for torchmd-net, but a 6x efficiency decrease is never a viable solution.
Possibly related: #27

RaulPPelaez:
TBH I do not really know what amount of the speedup is due to other improvements in lighting.
It makes sense to me that not testing is the main cause of the speedup simply because one is not going over a whole dataset every few epochs, which requires reading a bunch of stuff from disk, etc.
I understand the fast prototyping argument. But we are currently going against the lighting devs.
I will try to recreate the multiple-dataloader hack (looking at the docs/github issues I do not see any other workaround) so the functionality is still provided and I can measure if this is indeed the cause of the slowdown.

Well I guess I can also run the baseline without testing during training -.-


If that turns out to be the case then perhaps we can print a warning in the "test during training" case to explain that you are paying a high price for it hehe

giadefa:
I would drop it if it really makes such a huge difference. I don't see how
validation is not a problem and test costs 6 times.

RaulPPelaez:
Ok since I do not know yet how to reproduce the trick in latest Lighting I ran current main with and without testing during validation to compare:
image
Note that I was not trying to benchmark when I noticed this, I was checking correctness. Thus this train example must not be the most representative one (it does not fully occupy the GPU), but still...
Testing is carried out every 10 epochs in the orange line.
Roughly it is a x3 slowdown if you test every 10 epochs. Without testing I see a x3 speedup between Lightning 1.6.3 and 2.0.4.


I get testing being super slow because the way it is set up in this example it just goes over the whole dataset each time. Besides being just a lot of forward passes it requires reading the whole thing from disk.

With this I know believe that:
1. Updating Ligthing is very worth it.
2. We should keep the testing functionality for prototyping.
3. We should include a warning explaining that test-during-training can be expensive.

RaulPPelaez:
I managed to bring back the functionality.
As far as I can tell the only way to reproduce the test-during-train trick is to reload the dataloader every epoch.
This does not seem to have a profound effect in performance.
Sadly the performance when test-during-training is enabled does not look better than before.

I will test some more to make sure these performance numbers are not a fluke, but functionality-wise this should be ready to go!

torchmdnet/data.py
```diff
@@ -116,13 +116,16 @@ def mean(self):
     def std(self):
         return self._std
 
-    def _get_dataloader(self, dataset, stage, store_dataloader=True):
-        store_dataloader = (
-            store_dataloader and self.trainer.reload_dataloaders_every_n_epochs <= 0
```
RaulPPelaez, line 121: I removed this check here. 
By removing it the dataloaders are always stored.
AFAIK it was never actually used the check the project, so the behavior is unchanged. But I want to point it out in case I am missing something.

guillemsimeon:
I get that testing during training does only matter when you are prototyping, and in fact it helped in my case to stop training after few epochs when some hyperparameter change did not provide any improvement. If we are changing this, I have two suggestions:
- Report validation MAEs on both energies and forces on top of (separate) validation loss(es), which are MSEs. This can give a better sense of how the training is going.
- Even when no testing is performed during training, I would suggest in that case to provide test errors when the training is finished by default. Then people would not need to run inference separately on the test set to report performances.

guillemsimeon:
Another option is, when the training is finished, look for the checkpoint with lowest validation loss and run test with this one. I think this is what happens in MACE, for example.

RaulPPelaez:
Thanks for your insights!
I am convinced test-during-training has some useful cases, luckily I was able to cook it in again so I believe it is best to leave it there as an option at least for now. If you want to go go fast simply do not set the "test_interval" option.

> Another option is, when the training is finished, look for the checkpoint with lowest validation loss and run test with this one. I think this is what happens in MACE, for example.

AFAIK this is exactly what torchmd-train does:
https://github.com/torchmd/torchmd-net/blob/dca66796d00680a79a7b7c85d6704d30d15dc84c/torchmdnet/scripts/train.py#L179-L184

> Report validation MAEs

As you say, the values reported as "val_loss_y/dy" are mse_loss. I can add another reported value called "val_l1_loss_y/dy". It should have no effect whatsoever on performance.

giadefa:
reporting l2 and l1 loss on validation would be useful

RaulPPelaez:
I noticed there was no "on_test_epoch_end" member in LNNP. Meaning that the trainer.test line in train.py effectively just discarded all the work it did as far as I can tell. Also there is no logger attached to the test run, so nothing is written to disk about it.

I wrote one so that test losses are actually reported to the terminal like this:
image

RaulPPelaez:
This is ready to merge on my part.
@AntonioMirarchi, could you rerun some training with this PR to double check?
cc @raimis please review

torchmdnet/module.py
```diff
@@ -57,34 +57,95 @@ def configure_optimizers(self):
...
-            # validation step
-            return self.step(batch, mse_loss, "val")
-        # test step
-        return self.step(batch, l1_loss, "test")
+        # If args is not empty the first (and only) element is the dataloader_idx
+        # We want to test every number of epochs just for reporting, but this is not supported by Lightning.
+        # Instead, we trick it by providing two validation dataloaders and interpreting the second one as test.
+        # The dataloader takes care of sending the two sets only when the second one is needed.
+        is_val = len(args) == 0 or (len(args) > 0 and args[0] == 0)
+        step_type = (
```
stefdoerr, line 80: can you write this as a normal if/else please? It's less lines and more readable than a comprehension

torchmdnet/module.py
```diff
@@ -57,34 +57,95 @@ def configure_optimizers(self):
...
+                self.ema[stage][type][loss_name]
+                if loss_name in self.ema[stage][type]
+                else loss.detach()
+            )
+            loss = alpha * loss + (1 - alpha) * ema
+            self.ema[stage][type][loss_name] = loss.detach()
+        return loss
 
-    def step(self, batch, loss_fn, stage):
+    def _append_to_dict(self, d, k, v):
```
stefdoerr, line 130: check out https://docs.python.org/3/library/collections.html#collections.defaultdict

torchmdnet/module.py
```diff
@@ -163,64 +198,68 @@ def optimizer_step(self, *args, **kwargs):
...
-                self.trainer.reset_val_dataloader(self)
+    def _get_mean_loss_dict_for_type(self, type):
+        # Returns a list with the mean loss for each loss_fn for each stage (train, val, test)
+        # Parameters:
+        # type: either y, neg_dy or total
+        # Returns:
+        # A dict with an entry for each stage (train, val, test) with the mean loss for each loss_fn (e.g. mse_loss)
+        # The key for each entry is "stage_type_loss_fn"
+        assert self.losses is not None
+        self.mean_losses = {}
```
stefdoerr, line 209: why `self.`? I don't see this used anywhere else

RaulPPelaez: You are right!

torchmdnet/module.py
RaulPPelaez: Cool! did not know about this one

stefdoerr:
Test failed for some reason. Can you take a look?

RaulPPelaez:
It was just an overprotective numerical check. Just retriggering the CI has fixed it.

stefdoerr (approved):

guillemsimeon:
In the end is this instrinsically faster or was it just the effect of testing during training?

RaulPPelaez:
This small, not representative example, I am using is def faster beyond the test thing. 
Although every test I have run turns out faster, I would not expect real life runs to be much better. For instance, the ET-SPICE.yaml test, removing test-during-training, is just about 5% faster. YMMV.
I would expect to see improvements when epochs are really short, for instance.

cuicathy:
Hi, Thanks for the efforts to test-during-training. I wonder can we use test-during-training now? If yes, could you please give us an example? Thank you very much! 
P.S. I do not always need to test-during-training, but in some project for specific research purpose I have to...

RaulPPelaez:
Hi, The behavior is just like before this PR via the "test_interval" option. However, setting it to -1 will skip test during training.

Check out this doc page https://torchmd-net.readthedocs.io/en/latest/torchmd-train.html

PR #214, opened 2023-09-18
Make OutputModel aware of CUDA graph capturing.

RaulPPelaez:
The reduction in OutputModel requires knowing the number of batches (computed as batch.max()+1) to be able to produce a tensor with a per-batch property (a.i. the total energy). 
This operation is inherently blocking and thus cannot be called during CUDA graph capture.
 
One way to go around this limitation is to make the OutputModel store the number of batches each time it is called.
Then, if the module is called in capture mode (which can be detected with `torch.cuda.is_current_stream_capturing()`) the previously stored value is used instead.
This makes it so that the number of batches of a captured model cannot be larger than the one used in the call previous to capture. However, this limitation is proper to CUDA graphs and thus I do not consider it an additional limitation.

This PR implements this strategy, warning the user about this when they capture the model and raising if the model is being called without warming up first.
Possible alternatives:
1. Adding a new parameter to TorchMD_Net (like n_batches). I think this is more error prone.
2. Computing  and storing batch.max()+1 during capture without including it into the graph. I do not think this is possible sadly.


RaulPPelaez:
Welp, I was not counting on this one...
```shell
Python builtin <built-in function _cuda_isCurrentStreamCapturing> is currently not supported in Torchscript:
```

raimis:
I have an idea how to circument this

RaulPPelaez:
As a matter of fact this does not use nvcc, I think everything required is in cudatoolkit!

torchmdnet/models/output_modules.py
```diff
@@ -25,7 +80,22 @@ def pre_reduce(self, x, v, z, pos, batch):
...
-        return scatter(x, batch, dim=0, reduce=self.reduce_op)
+        is_capturing = (
+            x.is_cuda
+            and check_stream_capturing()
+        )
+        if not x.is_cuda or not is_capturing:
+            self.dim_size = int(batch.max().item() + 1)
+        if is_capturing:
+            assert (
+                self.dim_size == 0
```
raimis, line 91: Maybe
```python
self.dim_size > 0
```

torchmdnet/models/output_modules.py
```diff
@@ -6,16 +6,71 @@
...
+'''
+
+    # Create an inline extension
+    load_inline(
+        "is_stream_capturing",
+        cpp_sources=cpp_source,
+        functions=["is_stream_capturing"],
+        with_cuda=torch.cuda.is_available(),
+        extra_cflags=["-DWITH_CUDA"] if torch.cuda.is_available() else None,
+        verbose=True,
```
raimis, line 53: No need to be verbose

raimis: OK!

raimis (changes requested):
It works! For clarity, if you can move `check_stream_capturing` to a separate file, this is ready to be merged.

RaulPPelaez:
@raimis please review again so I can merge

raimis (approved):

Issue #215, opened 2023-09-19
conda-forge package


raimis:
We want easy integration with OpenMM via OpenMM-ML. This requires a conda-forge package, so the software stack can installed easily.

RaulPPelaez:
https://github.com/conda-forge/staged-recipes/pull/24021

RaulPPelaez:
It would make sense to use CUDA 12 for this package, but there are some dependencies preventing us from doing so.
In particular pytorch from conda-forge is not yet built for CUDA 12. I have been trying to help for some time but have not had luck, there is always some compilation issue https://github.com/conda-forge/pytorch-cpu-feedstock/pull/182

RaulPPelaez:
The main issue with CUDA is that the torch extensions in torchmd-net require nvcc at runtime (they are jit compiled) and currently nvcc installation is kind of borked in conda-forge because of the CUDA 12 migration.
OTOH conda-forge offers a simple way to get nvcc at conda-package-build time via the compiler('nvcc') directive.

One way to escape this issue is thus to compile the extensions during installation, so that nvcc is no longer a runtime dependency. 

I opened #216 to do that.

When it is merged I will create a new release (I would say 0.3) and package that.

raimis:
Yes, it looks like a good solution!

RaulPPelaez:
Closing as completed

PR #216, opened 2023-09-20
Compile torch extension in setup.py

RaulPPelaez:
Currently the neighbor extension requires having nvcc installed when importing it.

This PR makes it so that the extension is compiled when torchmdnet is pip installed. This removes nvcc as a runtime dependency, which solves the problem of packaging it in the upcoming conda-forge package #215

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -20,3 +20,4 @@ dependencies:
   - pytest
   - psutil
   - ninja
+  - gxx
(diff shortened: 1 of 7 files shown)


RaulPPelaez:
Ready to merge, @raimis please review.

environment.yml
```diff
@@ -20,3 +20,4 @@ dependencies:
   - pytest
   - psutil
   - ninja
+  - gxx
```
RaulPPelaez, line 23: I added this but it will not be a runtime dependency for the conda-package

raimis (approved):
Looks good!

PR #217, opened 2023-09-20
Allow to specify num_mols in OutputModel

raimis:
Alternative to #214 

Just specify the number of molecules to make CUDA Graphs work:
```python 
model = load_model(model_file, num_mols=1)
```


RaulPPelaez:
I am scared to add another parameter like this.
If one saves this model it will lock it to that num_mol.
On the other hand this parameter exists solely for the purpose of CUDA graphs, but it can be misleading to an user who might think this parameter _has_ to be set.

If we do this we have to document it really good.

raimis:
The solution in #214 is better.

PR #218, opened 2023-09-21
Use update instead of |= so Trainer is compatible with Python 3.8

RaulPPelaez:
The conda-forge package tries to build for Python 3.8, AFAIK there is only one operation that prevents the tests from running, so I changed it.


RaulPPelaez:
@raimis plase review

raimis (approved):

PR #223, opened 2023-09-26
Fix module.py

AntonioMirarchi:
- [x] `compute_losses` output return a dict with "y" and "neg_dy"; these 2 variables are initialized as float: 
	- if "derivative" is set to false or "neg_dy" not in batch, the neg_dy output is a float and .detach() function called in step give an error;
	- if "y" not in batch, "y" is a float (0.0) and the detach function called in step give an error; 
 According to the old version of the code I added a check for `neg_dy_weight` and `y_weight `
- [x] wandb error:
	- if you want to train with just "y" or "neg_dy", wandb should visualize only the relative metrics;
	- remove same metrics with different name (redundant), I modified `ModelCheckpoint()`;

### DIFF
--- a/torchmdnet/module.py
+++ b/torchmdnet/module.py
@@ -107,10 +107,7 @@ class LNNP(LightningModule):
         if "y" in batch:
             loss_y = loss_fn(y, batch.y)
             loss_y = self._update_loss_with_ema(stage, "y", loss_name, loss_y)
-        return {
-            "y": loss_y if torch.is_tensor(loss_y) else torch.zeros(()),
-            "neg_dy": loss_neg_y if torch.is_tensor(loss_neg_y) else torch.zeros(()),
-        }
+        return {"y": loss_y, "neg_dy": loss_neg_y}
 
     def _update_loss_with_ema(self, stage, type, loss_name, loss):
         # Update the loss using an exponential moving average when applicable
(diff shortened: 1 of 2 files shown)


RaulPPelaez:
You changed the checkpoint names, are you sure we want to do this?

torchmdnet/module.py
```diff
@@ -163,15 +163,21 @@ def step(self, batch, loss_fn_list, stage):
             batch.y = batch.y.unsqueeze(1)
         for loss_fn in loss_fn_list:
             step_losses = self._compute_losses(y, neg_dy, batch, loss_fn, stage)
+
+            loss_name = loss_fn.__name__
+
+            if self.hparams.neg_dy_weight > 0:
+                self.losses[stage]["neg_dy"][loss_name].append(
+                    step_losses["neg_dy"].detach()
+                )
```
RaulPPelaez, line 172: Is it ok to append only when it is non zero? As opposed to append zero

AntonioMirarchi: Essentially you don't care about the y loss metrics if `y_weight` is 0. If you don't append to the losses, then you are not going to have useless metrics in the logger (wandb) and in the `results_dict`.

torchmdnet/module.py
```diff
@@ -215,11 +221,6 @@ def on_validation_epoch_end(self):
             result_dict.update(self._get_mean_loss_dict_for_type("total"))
             result_dict.update(self._get_mean_loss_dict_for_type("y"))
```
RaulPPelaez, line 222: Does this not result in an error if you do not put 0 in the losses?

AntonioMirarchi: Not really, considering for example `y_weight=0.0`, this is what I get from `_get_mean_loss_dict_for_type` 

```
ANALYSIS OF TYPE total
stage train dict_keys(['mse_loss'])
stage val dict_keys(['l1_loss', 'mse_loss'])
stage test dict_keys([])
final:{'train_total_mse_loss': tensor(667.6912, device='cuda:0'), 'val_total_l1_loss': tensor(19.1803, device='cuda:0'), 'val_total_mse_loss': tensor(625.8098, device='cuda:0')}

ANALYSIS OF TYPE y
stage train dict_keys([])
stage val dict_keys([])
stage test dict_keys([])
final mean_losses:{}

ANALYSIS OF TYPE neg_dy
stage train dict_keys(['mse_loss'])
stage val dict_keys(['l1_loss', 'mse_loss'])
stage test dict_keys([])
final mean_losses:{'train_neg_dy_mse_loss': tensor(667.6912, device='cuda:0'), 'val_neg_dy_l1_loss': tensor(19.1803, device='cuda:0'), 'val_neg_dy_mse_loss': tensor(625.8098, device='cuda:0')}
```

All the dict in `self.losses `are default dict, so if a key is not present it return a empty list.

torchmdnet/module.py
RaulPPelaez: Please remove empty lines

RaulPPelaez:
Thanks @AntonioMirarchi !

PR #227, opened 2023-10-04
Unify torch extension compilation so that everything is compiled in setup.py

RaulPPelaez:
I wrote a unified mechanism to add TorchScript C++ extensions, which will be compiled in setup.py time and will be made available under the torchmdnet.extensions namespace.

I added the is_current_stream_capturing utility to it. This removes gcc and nvcc as runtime dependencies.


raimis (approved):
Looks good and clearer implementation

PR #229, opened 2023-10-09
Remove dependency on torch_scatter and torch_sparse

RaulPPelaez:
Removes dependency on torch_sparse by implementing message passing operations with manually written ones.

Removes dependency on torch_scatter by using torch.index_add and torch.scatter_reduce.


RaulPPelaez:
cc @guillemsimeon @AntonioMirarchi. Correctness tests are passing, but for this PR I had to reimplement the message passing of ET, T, and GN.

I do not want to merge it before a long train has been carried out to compare nothing broke. 
Can you recommend a good test yaml for this? Please post it here and I will run it with/without this PR.

guillemsimeon:
For the ET, the MD17 is fine. For the T, I would directly remove the architecture from TorchMD-NET, and I do not know any case where it has been used. For the GN, you can perhaps use the ones here https://arxiv.org/pdf/2212.07492.pdf

RaulPPelaez:
@giadefa, do you agree with removing the Transformer architecture altogether? Its less dev burden for sure...

guillemsimeon:
Raul, could you also train TensorNet on QM9 U0 with seed 1 as part of these tests? The yaml is in the repo.

RaulPPelaez:
I would say ET is correctly implemented:
image

RaulPPelaez:
TensorNet also looks ok:
image

RaulPPelaez:
GN also seems fine:
image
Not sure what makes GN be so much more accurate in terms of reproducibility than the other ones, though.

guillemsimeon (approved):
it looks good to me

PR #230, opened 2023-10-10
Update TensorNet-QM9.yaml

guillemsimeon:
I updated the optimal hyperparameters for TensorNet on QM9 to obtain the results of the camera-ready paper, which give an improvement over previous accuracy.

### DIFF
--- a/examples/TensorNet-QM9.yaml
+++ b/examples/TensorNet-QM9.yaml
@@ -1,7 +1,7 @@
 activation: silu
 aggr: add
 atom_filter: -1
-batch_size: 32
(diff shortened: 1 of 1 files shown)


RaulPPelaez:
You changed the seed, but I guess this is just a side effect, right? Like it should not be changing the accuracy noticeably. 
If this is the seed you used for the paper then its ok for reproducibility.

guillemsimeon:
Yes, the seed for the paper, the learning rate, and the batch size.

RaulPPelaez (approved):

PR #231, opened 2023-10-10
Log epoch real time in LNNP

RaulPPelaez:
This PR changes the LNNP module so that the real time since the start is logged each epoch.
Allows to track training time with the CSVLogger.

### DIFF
--- a/torchmdnet/module.py
+++ b/torchmdnet/module.py
@@ -2,10 +2,10 @@ from collections import defaultdict
 import torch
 from torch.optim import AdamW
 from torch.optim.lr_scheduler import ReduceLROnPlateau
-from torch.nn.functional import local_response_norm, mse_loss, l1_loss
+from torch.nn.functional import mse_loss, l1_loss
 from torch import Tensor
 from typing import Optional, Dict, Tuple
-
+import time
 from lightning import LightningModule
 from torchmdnet.models.model import create_model, load_model
 
@@ -36,6 +36,8 @@ class LNNP(LightningModule):
         # initialize loss collection
         self.losses = None
         self._reset_losses_dict()
+
+        self.tstart = time.time()
 
     def configure_optimizers(self):
         optimizer = AdamW(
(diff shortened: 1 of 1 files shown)


RaulPPelaez:
@AntonioMirarchi please review

AntonioMirarchi:
It works fine. This is the time column form the metrics.csv: 
```
0      4.779309
1      6.532793
2      8.925880
3     10.662247
4     13.062045
5     14.766587
6     17.175591
7     18.913141
8     21.361235
9     23.063034
10          NaN
Name: time, dtype: float64
```
The Nan on the last row is not clear to me

torchmdnet/module.py
```diff
@@ -215,6 +217,7 @@ def on_validation_epoch_end(self):
                 "epoch": float(self.current_epoch),
                 "lr": self.trainer.optimizers[0].param_groups[0]["lr"],
             }
+            result_dict["time"] = time.time() - self.tstart
```
AntonioMirarchi, line 220: maybe you can put directly above like for "epoch" and "lr"

RaulPPelaez:
I am only logging the time in `on_validation_epoch_end` and `on_test_epoch_end`. Perhaps there is some other place where data is logged?

RaulPPelaez:
Antonio can you give me a yaml to reproduce this NaN thing you see? I cannot trigger it.

RaulPPelaez:
I think the NaN Antonio reports comes from the test pass at the end of a train. It is considered a new "training"  and for some reason the first time has NaN.

RaulPPelaez:
Tried to change the logging of the epoch to integer, but this produces a warning:

```shell
 You called `self.log('epoch', ...)` in your `on_validation_epoch_end` but the value needs to be floating to be reduced. Converting it to torch.float32. You can silence this warning by converting the value to floating point yourself. If you don't intend to reduce the value (for instance when logging the global step or epoch) then you can use `self.logger.log_metrics({'epoch': ...})` instead.
```
There is some (really unconvincing imo) discussion on why one cannot log an integer https://github.com/Lightning-AI/pytorch-lightning/issues/18739
The alternative suggested in the warning itself does not work as one would expect, you get a new line in metrics.csv in which everything is empty except epoch.
cc @peastman

PR #233, opened 2023-10-11
Fix trying to detach a float number in LNNP

RaulPPelaez:
If the user trains with only energy or only forces in some situations the LNNP class will try to call detach on a float where it expects a tensor.
I changed the loss computation so it initializes the losses on both energy and force as a torch.tensor.

### DIFF
--- a/torchmdnet/module.py
+++ b/torchmdnet/module.py
@@ -97,7 +97,9 @@ class LNNP(LightningModule):
         # Returns:
         #   loss_y: loss for the predicted value
         #   loss_neg_y: loss for the predicted negative derivative
-        loss_y, loss_neg_y = 0.0, 0.0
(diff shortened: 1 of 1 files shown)


RaulPPelaez:
@raimis could you please review?

stefdoerr (approved):
I am not sure how optimized pytorch is but it might be slightly better to do `torch.zeros(1, device=self.device)`. In theory the above code will copy the value to the GPU while the other one might set it directly on the device. It's *extremely* insignificant performance-wise for a single number but I assume there is no reason not to do it :P Up to you

stefdoerr:
Sorry scratch that, torch.zeros makes an array, torch.tensor makes a single float. Ignore my comment

PR #235, opened 2023-10-11
Preload small datasets to memory in Custom dataset

RaulPPelaez:
While replicating results from the [torchmd-protein-thermodynamics repository](https://github.com/torchmd/torchmd-protein-thermodynamics), I experienced sluggish training speeds and low GPU usage (meaning sitting at 0% and briefly going to 100% at each iteration) using the following configuration file:
```yaml
activation: tanh
batch_size: 1024
inference_batch_size: 1024
dataset: Custom
coord_files: "chignolin_ca_coords.npy"
embed_files: "chignolin_ca_embeddings.npy"
force_files: "chignolin_ca_deltaforces.npy"
cutoff_upper: 12.0
cutoff_lower: 3.0
derivative: true
early_stopping_patience: 30
embedding_dimension: 128
lr: 0.0005
lr_factor: 0.8
lr_min: 1.0e-06
lr_patience: 10
lr_warmup_steps: 0
model: graph-network
neighbor_embedding: false
ngpus: -1
num_epochs: 100
num_layers: 4
num_nodes: 1
num_rbf: 18
num_workers: 4
rbf_type: expnorm
save_interval: 2
seed: 1
test_interval: 2
test_size: 0.1
trainable_rbf: true
val_size: 0.05
weight_decay: 0.0
```
The referenced files take approx 300mb.
Playing around with num_workers and batch size did not help.
Upon investigation, the issue arose from the Custom dataset's I/O-bound get method, which reads from disk every time it is invoked, causing the low GPU usage and slow training.

To resolve this, I implemented a preloading feature that loads the complete dataset into system memory if its size is below a user-configurable threshold, set by default at 1GB. The data is stored as PyTorch tensors, facilitating compatibility with multi-threaded data loaders (num_workers). Notably, this approach does not inflate RAM usage when increasing the number of workers.

This optimization led to a x20 speedup in training time for this specific setup.

OTOH I tweaked the DataLoader options a bit.

### DIFF
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -10,23 +10,26 @@ import h5py
 @mark.parametrize("energy", [True, False])
 @mark.parametrize("forces", [True, False])
 @mark.parametrize("num_files", [1, 3])
-@mark.parametrize("preload", [False, True])
-def test_custom(energy, forces, num_files, preload, tmpdir, num_samples=100):
+@mark.parametrize(("preload", "read_as_hdf5"), [(True, None), (False, "test.hdf5")])
+def test_custom(energy, forces, num_files, preload, read_as_hdf5, tmpdir, num_samples=100):
+    if energy == False and isinstance(read_as_hdf5, str):
+        pytest.skip("HDF5 format requires energies to be present in the dataset.")
+    np.random.seed(0)
     # set up necessary files
     for i in range(num_files):
         np.save(
-            join(tmpdir, f"coords_{i}.npy"), np.random.normal(size=(num_samples, 5, 3))
+            join(tmpdir, f"coords_{i}.npy"), np.random.normal(size=(num_samples, 5, 3)).astype(np.float32)
         )
         np.save(join(tmpdir, f"embed_{i}.npy"), np.random.randint(0, 100, size=(5)))
         if energy:
             np.save(
                 join(tmpdir, f"energy_{i}.npy"),
-                np.random.uniform(size=(num_samples, 1)),
+                np.random.uniform(size=(num_samples, 1)).astype(np.float32),
             )
         if forces:
             np.save(
                 join(tmpdir, f"forces_{i}.npy"),
-                np.random.normal(size=(num_samples, 5, 3)),
+                np.random.normal(size=(num_samples, 5, 3)).astype(np.float32),
             )
 
     # load data and test Custom dataset
@@ -38,12 +41,15 @@ def test_custom(energy, forces, num_files, preload, tmpdir, num_samples=100):
             )
         return
 
+    if read_as_hdf5 is not None:
+        read_as_hdf5 = join(tmpdir, read_as_hdf5)
     data = Custom(
         coordglob=join(tmpdir, "coords*"),
         embedglob=join(tmpdir, "embed*"),
         energyglob=join(tmpdir, "energy*") if energy else None,
         forceglob=join(tmpdir, "forces*") if forces else None,
         preload_memory_limit=256 if preload else 0,
+        read_as_hdf5=read_as_hdf5,
     )
 
     assert len(data) == num_samples * num_files, "Number of samples does not match"
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -10,7 +10,8 @@ import h5py
 @mark.parametrize("energy", [True, False])
 @mark.parametrize("forces", [True, False])
 @mark.parametrize("num_files", [1, 3])
-def test_custom(energy, forces, num_files, tmpdir, num_samples=100):
+@mark.parametrize("preload", [False, True])
+def test_custom(energy, forces, num_files, preload, tmpdir, num_samples=100):
     # set up necessary files
     for i in range(num_files):
         np.save(
@@ -42,6 +43,7 @@ def test_custom(energy, forces, num_files, tmpdir, num_samples=100):
         embedglob=join(tmpdir, "embed*"),
         energyglob=join(tmpdir, "energy*") if energy else None,
         forceglob=join(tmpdir, "forces*") if forces else None,
+        preload_memory_limit=256 if preload else 0,
     )
 
     assert len(data) == num_samples * num_files, "Number of samples does not match"
@@ -52,6 +54,20 @@ def test_custom(energy, forces, num_files, tmpdir, num_samples=100):
         assert hasattr(sample, "y"), "Sample doesn't contain energy"
     if forces:
         assert hasattr(sample, "neg_dy"), "Sample doesn't contain forces"
+
+    # Assert sample has the correct values
+
+    # get the reference values from coords_0.npy and embed_0.npy
+    ref_pos = np.load(join(tmpdir, "coords_0.npy"))[0]
+    ref_z = np.load(join(tmpdir, "embed_0.npy"))
+    assert np.allclose(sample.pos, ref_pos), "Sample has incorrect coords"
+    assert np.allclose(sample.z, ref_z), "Sample has incorrect atom numbers"
+    if energy:
+        ref_y = np.load(join(tmpdir, "energy_0.npy"))[0] if energy else None
+        assert np.allclose(sample.y, ref_y), "Sample has incorrect energy"
+    if forces:
+        ref_neg_dy = np.load(join(tmpdir, "forces_0.npy"))[0] if forces else None
+        assert np.allclose(sample.neg_dy, ref_neg_dy), "Sample has incorrect forces"
 
 
 def test_hdf5_multiprocessing(tmpdir, num_entries=100):
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -5,8 +5,9 @@ from os.path import join
 import numpy as np
 import psutil
 from torchmdnet.datasets import Custom, HDF5
+from torchmdnet.utils import write_as_hdf5
 import h5py
-
+import glob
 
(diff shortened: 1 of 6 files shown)


peastman:
The Custom dataset class is incredibly inefficient.  It reloads the whole file every time `get()` is called to retrieve a single sample.  Some intelligent caching would help.  But a much better choice is to use the HDF5 dataset class.  It is far more efficient.

peastman:
In fact, possibly we should just make Custom create a temporary HDF5 file on startup and then load from it.

RaulPPelaez:
I really like this idea Peter, thanks!
I do not like Custom at all either, but it offers some convenience/simplicity that makes people choose it so I think it is worth improving.
I implemented so that the user can instruct Custom to use HDF5 under the hood, lets see how it goes.

torchmdnet/datasets/custom.py
```diff
@@ -2,6 +2,34 @@
...
+            group = f.create_group(str(i))
+            num_samples = coord_data.shape[0]
+            group["pos"] = coord_data
+            group["types"] = np.tile(embed_data, (num_samples, 1))
+            if "y" in files:
+                energy_data = np.load(files["y"][i])
+                group["energy"] = energy_data
+            if "neg_dy" in files:
+                force_data = np.load(files["neg_dy"][i])
+                group["forces"] = force_data
```
AntonioMirarchi, line 32: I think the better way to create a hdf5 file (especially for the datasets) is: 
```python 
f = h5py.File(hdf5_dataset, "w")
for i in range(len(files["pos"])):
    # Create a group for each file
    group = f.create_group(str(i))
    coord_data = np.load(files["pos"][i], mmap_mode="r")
    embed_data = np.load(files["z"][i], mmap_mode="r").astype(int)
    num_samples = coord_data.shape[0]
    group.create_dataset("pos", data=coords_data)
    group.create_dataset("types", data=np.tile(embed_data, (num_samples, 1)))
    if "y" in files:
        energy_data = np.load(files["y"][i], mmap_mode="r")
        group.create_dataset("energy", data=energy_data)
    if "neg_dy" in files:
        force_data = np.load(files["neg_dy"][i], mmap_mode="r")
        group.create_dataset("forces", data=force_data)
f.close()
```
I also added `mmap_mode="r"` when you call numpy load. These are just tips, feel free to skip it if you think are not useful.

RaulPPelaez: mmap_mode makes sense, as the dataset might be super big.
Why is group.create_dataset("something",... better than group["something"] = ?

torchmdnet/datasets/custom.py
```diff
@@ -30,87 +60,161 @@ class Custom(Dataset):
...
+            write_as_hdf5(
+                self.files,
+                hdf5_file,
+            )
+            self.hdf5_dataset = HDF5(hdf5_file)
+        total_data_size = self._initialize_index()
+        print("Combined dataset size {}".format(len(self.index)))
+        # If the dataset is small enough, load it whole into CPU memory
+        data_size_limit = preload_memory_limit * 1024 * 1024
+        self.cache = False
```
AntonioMirarchi, line 116: Check this, I think you meant `self.cached`  but you already defined it in line 104.

torchmdnet/datasets/custom.py
AntonioMirarchi: It's just the "proper way", in theory you should be able to move across path in the hdf5 file (it's not this case) . For example if you use "create_dataset" you can retrieve the pos dataset using:
```python 
f = h5py.File(hdf5_dataset, "r")
f["0/pos"]
```
While if you use the "dictionary way" you get this error: 
`KeyError: "Unable to open object (object 'pos' doesn't exist)"`

giadefa:
h5py are slower than mmap arrays. Maybe I am missing the narrative but
why are we doing something different from what we are already doing in
other dataloaders?


g

torchmdnet/datasets/custom.py
AntonioMirarchi: I have always used that way, assuming was the proper one.  [reference](https://docs.h5py.org/en/stable/high/dataset.html)

RaulPPelaez: Convinced! Thanks.
I am not sure repeating the type data is necessary, but I found no way around it:
    group.create_dataset("types", data=np.tile(embed_data, (num_samples, 1)))
Can you think of something?

RaulPPelaez:
This PR exists because Custom was calling np.load() at each call of get(). Even with mmap mode this was really slowing down trainings (it is I/O bound in the end...).
I changed it so that:
1- If the dataset is small enough just load it all into RAM
2- Otherwise store the references to the mmap arrays instead of reloading the file each time.
Alternatively, I added an option to transform the Custom files to HDF5, which seems to be just a little bit slower than mmap.
I went ahead and also implemented the same load to RAM functionality in HDF5.

torchmdnet/datasets/custom.py
```diff
@@ -30,87 +60,160 @@ class Custom(Dataset):
...
+            torch.from_numpy(np.load(f)) for f in self.files["pos"]
+        ]
+        self.stored_data["z"] = [
+            torch.from_numpy(np.load(f).astype(int))
+            .unsqueeze(0)
+            .expand(self.stored_data["pos"][i].shape[0], -1)
+            for i, f in enumerate(self.files["z"])
+        ]
+        if self.has_energies:
+            self.stored_data["y"] = [
```
AntonioMirarchi, line 142: Is `torch.from_numpy()` what you really need? Take into account that if you modify the tensor then these changes will be reflected in the ndarray and vice versa. Moreover, the returned tensor is not resizable. I would say `torch.tensor()` is fine.  I think you can use also here the `mmap_mode="r"`, it will work only with `torch.tensor()`

RaulPPelaez: But I do not want to mmap these ones, I want to load them into memory, they will never be resized or mutated in any way, only accesed.

torchmdnet/data.py
```diff
@@ -55,17 +55,23 @@ def setup(self, stage):
                     self.hparams["embed_files"],
                     self.hparams["energy_files"],
                     self.hparams["force_files"],
+                    self.hparams["dataset_preload_limit"],
+                    self.hparams["custom_read_as_hdf5"],
```
AntonioMirarchi, line 59: remove arg that is not implemented anymore

AntonioMirarchi: just `custom_read_as_hdf5`

RaulPPelaez:
This is ready. @stefdoerr could you please review?

tests/test_datasets.py
```diff
@@ -1,32 +1,49 @@
...
+import glob
 
-
-@mark.parametrize("energy", [True, False])
-@mark.parametrize("forces", [True, False])
-@mark.parametrize("num_files", [1, 3])
-def test_custom(energy, forces, num_files, tmpdir, num_samples=100):
-    # set up necessary files
+def write_sample_npy_files(energy, forces, tmpdir, num_files):
+        # set up necessary files
```
stefdoerr, line 13: is this an indentation error?

torchmdnet/data.py
```diff
@@ -55,17 +55,22 @@ def setup(self, stage):
                     self.hparams["embed_files"],
                     self.hparams["energy_files"],
                     self.hparams["force_files"],
+                    self.hparams["dataset_preload_limit"],
                 )
             else:
                 dataset_arg = {}
                 if self.hparams["dataset_arg"] is not None:
                     dataset_arg = self.hparams["dataset_arg"]
+                if self.hparams["dataset"] == "HD5F":
```
stefdoerr, line 64: HDF5 you mean? is this a typo?

torchmdnet/datasets/custom.py
```diff
@@ -30,87 +34,150 @@ class Custom(Dataset):
...
+                f"Number of coordinate files {len(self.files['pos'])} "
+                f"does not match number of force files {len(self.files['neg_dy'])}."
             )
-
-        print("Number of files: ", len(self.coordfiles))
-
+        print("Number of files: ", len(self.files["pos"]))
+        self.cached = False
+        total_data_size = self._initialize_index()
+        print("Combined dataset size {}".format(len(self.index)))
```
stefdoerr, line 79: Ideally don't use `.format`. f-strings are much more readable and modern

```py
print(f"Combined dataset size {len(self.index))}")
```

torchmdnet/datasets/custom.py
```diff
@@ -30,87 +34,150 @@ class Custom(Dataset):
...
+        # If the dataset is small enough, load it whole into CPU memory
+        data_size_limit = preload_memory_limit * 1024 * 1024
+        if total_data_size < data_size_limit:
+            self.cached = True
+            print(
+                "Preloading Custom dataset (of size {:.2f} MB)".format(
+                    total_data_size / 1024**2
```
stefdoerr, line 86:
```py
f"Preloading Custom dataset (of size {total_data_size / (1024**2):.2f} MB)"
```
works also with floating point formatting if you add the number before the `:`

stefdoerr:
I think @raimis should review this since he worked on mmaps before. I am not too qualified, so I just commented on style and minor bugs

RaulPPelaez:
I would like to merge this one before next release, @raimis could you take a look? Thanks

guillemsimeon:
is this ready to be shipped?

RaulPPelaez:
Yes!

guillemsimeon (approved):
perf

PR #236, opened 2023-10-18
Fix Tensornet accuracy bug

RaulPPelaez:
In CUDA graph mode in tensornet static shapes require processing invalid neighbor pairs (which are passed down as (-1,-1) pairs).
By blaming all extra neighbor pairs to the first particle the overall accuracy could be slightly affected in some situations.
I fix this by instead blaming the extra pairs to a ghost atom instead.


guillemsimeon (approved):
Approve

PR #237, opened 2023-10-25
Update README with conda-forge installation

RaulPPelaez:


RaulPPelaez:
Yes!

RaulPPelaez:
What option are you referring to?

guillemsimeon:
The static shapes thing for example, which I am seeing defaults to True in TensorNet, and I understand this needs to be done to use CUDA graphs. But don't take very seriously my thoughts, I might be wrong, you decide.

RaulPPelaez:
Yeah... there is no option to tune that from outside, its always true. 
There is no other performance-only option, so I did not want to include it just for that. Additionally, for that particular option, there is really no good heuristic AFAIK. I tried to enable it only if cuda graph capture is enabled but torch really dislikes these kind of model behavior change and I run into issues. So I just left it as True.

At some point it would be interesting to add a set of performance-only options, but I think this is not the time.

guillemsimeon (approved):
Great

PR #238, opened 2023-11-02
Add support for charged molecules in TensorNet (total charge)

guillemsimeon:
Training using total charges should be possible by just setting the charge hyperparameter to True in the yaml file.

### DIFF
--- a/torchmdnet/models/tensornet.py
+++ b/torchmdnet/models/tensornet.py
@@ -216,6 +216,7 @@ class TensorNet(nn.Module):
         if self.static_shapes:
             mask = (edge_index[0] < 0).unsqueeze(0).expand_as(edge_index)
             zp = torch.cat((z, torch.zeros(1, device=z.device, dtype=z.dtype)), dim=0)
+            # Total charge q is a molecule-wise property. We transform it into an atom-wise property, with all atoms belonging to the same molecule being assigned the same charge q
             if q is None:
                 q = torch.zeros_like(zp, device=z.device, dtype=z.dtype)
             else:
@@ -233,6 +234,7 @@ class TensorNet(nn.Module):
         # I avoid dividing by zero by setting the weight of self edges and self loops to 1
         edge_vec = edge_vec / edge_weight.masked_fill(mask, 1).unsqueeze(1)
         X = self.tensor_embedding(zp, edge_index, edge_weight, edge_vec, edge_attr)
+        # Total charge q is a molecule-wise property. We transform it into an atom-wise property, with all atoms belonging to the same molecule being assigned the same charge q
         if q is None:
             q = torch.zeros_like(z, device=z.device, dtype=z.dtype)
         else:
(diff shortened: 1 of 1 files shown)


torchmdnet/models/tensornet.py
```diff
@@ -217,6 +217,12 @@ def forward(
             mask = (edge_index[0] < 0).unsqueeze(0).expand_as(edge_index)
             zp = torch.cat((z, torch.zeros(1, device=z.device, dtype=z.dtype)), dim=0)
             # I trick the model into thinking that the masked edges pertain to the extra atom
+            if q is None:
+                q = 0
+            else:
+                batchp = torch.cat((batch, batch[-1] + 1), dim=0)
+                qp = torch.cat((q, torch.zeros(1, device=q.device, dtype=q.dtype)), dim=0)
+                q = qp[batchp][...,None,None,None]
```
RaulPPelaez, line 225: Put this above the comment so is not a lie.

guillemsimeon: whoops, true

RaulPPelaez:
There are some error:
```shell
FAILED tests/test_model.py::test_torchscript[cpu-tensornet] - RuntimeError: 
Variable 'q' previously had type Optional[Tensor] but is now being assigned to a value of type int
:
  File "/usr/share/miniconda3/envs/torchmd-net/lib/python3.10/site-packages/torchmdnet/models/tensornet.py", line 221
            # I trick the model into thinking that the masked edges pertain to the extra atom
            if q is None:
                q = 0
                ~ <--- HERE
            else:
                batchp = torch.cat((batch, batch[-1] + 1), dim=0)
```
Default q to something like 
```python
q = torch.zeros(1, device=z.device, dtype=pos.dtype)
```

torchmdnet/models/tensornet.py
```diff
@@ -227,8 +234,13 @@ def forward(
         # I avoid dividing by zero by setting the weight of self edges and self loops to 1
         edge_vec = edge_vec / edge_weight.masked_fill(mask, 1).unsqueeze(1)
         X = self.tensor_embedding(zp, edge_index, edge_weight, edge_vec, edge_attr)
+        # Total charge q is a molecule-wise property. We transform it into an atom-wise property, with all atoms belonging to the same molecule being assigned the same charge q
+        if q is None:
+            q = torch.zeros_like(z, device=z.device, dtype=z.dtype)
+        else:
+            q = q[batch]
```
RaulPPelaez, line 241: q is never None here, right?

RaulPPelaez: Also you already set up q in 225

guillemsimeon: I understand that if charge = False in the yaml, q will be None. Or it really never goes through the representation model then?
And the q in 225 is when static shapes is True.

RaulPPelaez: in line 220 you check if q is None, then both branches define q, I do not see how q can be None in line 238 then.
I also do not understand why you need q = q[batch] in line 241. Wont it fail because you did not store batchp?
As far as I can tell these lines should just go.

RaulPPelaez:
Looks good and GPU tests pass. Lets merge. Thanks @guillemsimeon !

RaulPPelaez (approved):

Issue #198, opened 2023-06-30
Support higher dimension outputs in TorchMD-Net


RaulPPelaez:
Right now the TorchMD_Net module is tied to the idea that only one thing (and maybe its derivative) is returned:
https://github.com/torchmd/torchmd-net/blob/a11684710d6ba22b29acbf709bfbc05b93f3402d/torchmdnet/models/model.py#L300

With TensorNet it might be interesting to have more outputs. Say a scalar (energies), minus its derivative (forces) and some tensor feature. 

I would like to discuss how to include this functionality while leaving the current TorchMD_Net interface as unmodified as possible.

A way to do this would be:

- Change the return type of TorchMD_Net to be an Union[Tensor, List[Tensor]]. derivative, if true, only diffs the first tensor.
https://github.com/torchmd/torchmd-net/blob/a11684710d6ba22b29acbf709bfbc05b93f3402d/torchmdnet/models/model.py#L240
  The consequences of this is a possible TorchScript nightmare
- Change the output models to take in a List[Tensor] for v. For instance here:
https://github.com/torchmd/torchmd-net/blob/a11684710d6ba22b29acbf709bfbc05b93f3402d/torchmdnet/models/output_modules.py#L168
 
cc @guillemsimeon

peastman:
This feature is needed for some things I want to work on.  It's also closely related to #26.

Here's my particular situation.  I want the model to predict charges for every atom.  Those charges will get factored into computing a Coulomb energy term.  This requires both multiple outputs and multiple loss terms.  Training will likely happen in multiple stages.  For example, a possible protocol would be

Stage 1. Use fixed charges in computing the Coulomb energy.  There are two loss terms: one that matches the output energy to the value in the dataset, and one that matches the atomic charges to values found in the dataset.  That means the two output heads are being trained independently, though of course much of the model is shared between them.

Stage 2. Use the predicted charges for computing the Coulomb energy.  The same two loss terms are used.  The predicted charges now affect both terms.

Stage 3. Drop the loss term for the charges, and fine tune the model based only on energy.

Once the model is trained, we will often only be interested in the energy output, which is used for running simulations.  But we might sometimes be interested in the predicted charges too.  For example, they could be used for parametrizing a molecule for a conventional MD simulation.

How could this be implemented?

giadefa:
On Thu, Aug 31, 2023 at 10:44 PM Peter Eastman ***@***.***>
wrote:

[...]

PR #239, opened 2023-11-03
[WIP] Support arbitrary outputs in TorchMD_Net

RaulPPelaez:
Following the discussion in https://github.com/torchmd/torchmd-net/issues/198 this PR attempts to give TorchMD_Net the ability to return more than one output ("y") and its derivative ("neg_dy").

This PR is still a draft as I am trying to figure out the final design.

#### This PR introduces user-facing breaking changes:
- It changes some names in the configuration file (for instance Scalar is no longer a thing). Although a conversion could be made when processing the configuration.
- The Datasets must provide "energy", "force" instead of "y", "neg_dy".
- TorchMD_Net is expected to compute always at least energy, instead of a generic label called "y". Maybe I am missing some usecases here, so we will see...

#### New design proposed for the outputs of the model:
- TorchMD_Net is composed of a representation model + an arbitrary number of heads stacked sequentially.
- There is no distinction between a Prior and what used to be an OutputModel, they are all Heads now.
- The EnergyHead is always the first one and the ForceHead the last (if derivative=True)
- There is some level of customization akin to the Heads for computing the loss of each output and reducing the total loss.
- The user provides a list of weights (like y_weight, neg_dy_weight now) for each model output that should be considered for the loss computation.

This is the BaseHead interface I propose:

```python
class BaseHead(nn.Module):
    def __init__(self, dtype=torch.float32):
        super(BaseHead, self).__init__()
        self.dtype = dtype

    def reset_parameters(self):
        pass

    def per_point(self, point_features, results, z, pos, batch, extra_args):
        return point_features, results

    def per_sample(self, point_features, results, z, pos, batch, extra_args):
        return point_features, results
```
Where the forward call of TorchMD_Net would go like this:
```python
        results = {}
        point_features = self.representation_model(z, pos, batch, q=q, s=s)
        for head in self.head_list:
            point_features, results = head.per_point(point_features, results, z=z, pos=pos, batch=batch, extra_args=extra_args)
        for head in self.head_list:
            point_features, results = head.per_sample(point_features, results, z=z, pos=pos, batch=batch, extra_args=extra_args)
```
Each head is free to add a new key to result, modify the point_features or the contents of result (i.e add to the energy). For instance, the EnergyHead:
```python
class EnergyHead(BaseHead):
    def __init__(self,
                 hidden_channels,
                 activation="silu",
                 dtype=torch.float32):
        super(EnergyHead, self).__init__(dtype=dtype)
        act_class = act_class_mapping[activation]
        self.output_network = nn.Sequential(
            nn.Linear(hidden_channels, hidden_channels // 2, dtype=dtype),
            act_class(),
            nn.Linear(hidden_channels // 2, 1, dtype=dtype),
        )
        self.reset_parameters()

    def reset_parameters(self):
        nn.init.xavier_uniform_(self.output_network[0].weight)
        self.output_network[0].bias.data.fill_(0)
        nn.init.xavier_uniform_(self.output_network[2].weight)
        self.output_network[2].bias.data.fill_(0)

    def per_point(self, point_features, results, z, pos, batch, extra_args):
        results["energy"] = self.output_network(point_features)
        return point_features, results

    def per_sample(self, point_features, results, z, pos, batch, extra_args):
        results["energy"] = scatter(results["energy"], batch, dim=0)
        return point_features, results
```

There are some challenges I have still to deal with:
- Not sure how happy TorchScript is going to be with this.
- Not sure ho the user should specify a list of predefined heads. Perhaps something like an option 
```head_list: energy_head, coulomb_prior, some_other_prior, charge_head, some_charge_prior, force_head```

#### Tasks:
- [x] Adapt TorchMD_Net
- [ ] Reintroduce support for mean and std
- - [ ] Make Equivariant versions of the heads for ET.
- [x] Adapt LNNP
- [ ] Adapt Datasets
- [ ] Make priors into heads
- [ ] Generalize the loss computation
- [ ] Handle user input
- [ ] Update tests

### DIFF
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -313,9 +313,6 @@ class BaseHead(nn.Module):
     def per_point(self, point_features, results, z, pos, batch, extra_args):
         return point_features, results
 
-    def reduce(self, point_features, results, z, pos, batch, extra_args):
-        return point_features, results
-
     def per_sample(self, point_features, results, z, pos, batch, extra_args):
         return point_features, results
 

### CODE REVIEW

**Comment by giadefa** [comment]
Why are we changing these things?
We agreed on them a while back.

g

**Comment by peastman** [comment]
I was thinking of something a bit more generic than this.  You can define an arbitrary set of output heads and loss terms.  I imagine the description in the config file looking something like this.

```yaml
output_heads:
  - scalar:
    name: energy
  - coulomb  # the Coulomb head is hardcoded to output a scalar "energy" and a vector "charges"
losses:
  - l2
    output: energy  # since multiple heads have "energy" outputs, they get summed before computing the loss
    dataset_field: y
    weight: 1.0
  - gradient_l2
    output: energy
    dataset_field: neg_dy
    weight: 0.1
  - l2
    output: charges
    dataset_field: mbis_charges
    weight: 0.1
```

The configuration for a totally different sort of model might look like this.

```yaml
output_heads:
  - scalar
    name: solubility
losses:
  - l2
    output: solubility
    dataset_field: solubility
    # if weight is omitted, it defaults to 1
```

peastman:
Is it ok if I try implementing the design described above?

giadefa:
We have already an implementation of what I think it's what you need, so
maybe wait that Raul finds out what is that we are already doing.

G

PR #240, opened 2023-11-20
Mark Transformer as deprecated

RaulPPelaez:
The Transformer architecture is not being used anymore and it is becoming an unnecessary development burden.

For the moment, in this PR I am marking it as deprecated to be removed in a future release.

Please let me know if you think this is a bad idea or want to make a case for the Transformer.


guillemsimeon:
I would remove it. And + it is the only model that does not have an appropriate citation/justification since to the best of my knowledge has not been used.

RaulPPelaez:
I am weary of straight-up removing it now in case there is someone using it that we are not aware of -.-

PR #241, opened 2023-11-20
Add missing parse_args from previous bad merge

RaulPPelaez:


guillemsimeon (approved):
ups

RaulPPelaez:
I added a CI job to test that torchmd-train --help succeeds so this does not happen again :P

PR #246, opened 2023-12-17
Update tensornet.py

guillemsimeon:
update ref

### DIFF
--- a/torchmdnet/models/tensornet.py
+++ b/torchmdnet/models/tensornet.py
@@ -56,8 +56,9 @@ def tensor_norm(tensor):
     return (tensor**2).sum((-2, -1))
 
 class TensorNet(nn.Module):
-    r"""TensorNet's architecture.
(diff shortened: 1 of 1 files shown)


RaulPPelaez:
@guillemsimeon pls, update this reference here:

https://github.com/torchmd/torchmd-net/blob/af1aa2342e476a5af7eff9a1432b27d0fb94759c/docs/source/index.rst?plain=1#L45-L52

and here too:

https://github.com/torchmd/torchmd-net/blob/af1aa2342e476a5af7eff9a1432b27d0fb94759c/README.md?plain=1#L148-L159

guillemsimeon:
I would wait to update this reference when the proceedings of the NeurIPS are out, which would be the correct citation. For now, the preprint is fine. What other people do is they link the paper to the openreview version: https://openreview.net/pdf?id=BEHlPdBZ2e , but I am not sure how this can be incorporated in bib format.

RaulPPelaez:
Ok lets leave it like this for the moment then.

Issue #249, opened 2024-01-08
Errors installing


peastman:
I'm trying to install with the commands in the README, but so far without success:

```
mamba env create -f environment.yml
conda activate torchmd-net
mamba install cuda-nvcc cuda-libraries-dev cuda-version "gxx<12" pytorch=*=*cuda*
pip install -e .
```

The third line, to install the CUDA libraries, fails with the error

```
Could not solve for environment specs
The following packages are incompatible
├─ cuda-libraries-dev is installable with the potential options
│  ├─ cuda-libraries-dev 12.0.0 would require
│  │  └─ cuda-cccl 12.0.90.* , which requires
│  │     └─ cuda-version >=12.0.0,<12.1.0a0 , which requires
│  │        └─ cudatoolkit 12.0|12.0.* , which can be installed;
│  ├─ cuda-libraries-dev 12.1.1 would require
│  │  └─ libnvjpeg-dev 12.2.0.2.* , which requires
│  │     └─ cuda-version >=12.1,<12.2.0a0 , which requires
│  │        └─ cudatoolkit 12.1|12.1.* , which can be installed;
│  └─ cuda-libraries-dev 12.2.2 would require
│     └─ libnvjpeg-dev 12.2.2.4.* , which requires
│        └─ cuda-version >=12.2,<12.3.0a0 , which requires
│           └─ cudatoolkit 12.2|12.2.* , which can be installed;
└─ cudatoolkit is not installable because it conflicts with any installable versions previously reported.
```

The first command, to build the environment, installed CUDA 11.8:

```
cuda-version              11.8                 h70ddcb2_2    conda-forge
cudatoolkit               11.8.0              h4ba93d1_12    conda-forge
```

I don't understand exactly what's going on or where those dependencies are coming from.  Starting with CUDA 12, there's no longer any such package as `cudatoolkit`.  It's been broken up into separate packages for each library.  But something expects it to exist for 12.x.

RaulPPelaez:
Ugh! Thanks for checking Peter.

I can see the issue, when installing the environment mamba decided to install cuda 11.8, which brings cudatoolkit. Then when you try to install cuda-nvcc, which forces cuda-version to >=12, it will try to "update" cudatoolkit to >=12, which does not exists. IMO this behavior should be considered a bug in conda-forge, I do not understand why cudatoolkit>12 is not a metapackage with all the new cuda-* packages.

[...]

PR #250, opened 2024-01-09
Update installation instructions regarding CUDA

RaulPPelaez:
cc @peastman 
Follow up from #249

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,55 +29,7 @@ We recommend using [Mamba](https://github.com/conda-forge/miniforge/#mambaforge)
 
 ### Install from source  
 
-1. Clone the repository:
-    ```shell
-    git clone https://github.com/torchmd/torchmd-net.git
-    cd torchmd-net
-    ```
-
-2. Install the dependencies in environment.yml. You can do it via pip, but we recommend [Mambaforge](https://github.com/conda-forge/miniforge/#mambaforge) instead:  
-
-    ```shell
-    mamba env create -f environment.yml
-    mamba activate torchmd-net
-    ```
-
-3. CUDA enabled installation  
-   You can skip this section if you only need a CPU installation.
-   
-   The conda-forge channel [changed the way to install CUDA from versions 12 and above](https://github.com/conda-forge/conda-forge.github.io/issues/1963), thus the following instructions depend on whether you need CUDA < 12. If you have a GPU available, conda-forge probably installed the CUDA runtime (not the developer tools) on your system already, you can check with mamba:
-   
-   ```shell
-   mamba list | grep cuda
-   ```
-   Or by asking pytorch:
-   ```shell
-   python -c "import torch; print(torch.version.cuda)"
-   ```
-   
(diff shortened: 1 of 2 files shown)


peastman:
I would rather have it install CUDA 12 versions of everything.  I tried

```
CONDA_OVERRIDE_CUDA="12" mamba env create -f environment.yml
```

but it still installs 11.8.  I'm not sure where it's getting that from.  Presumably it must be required by one of the dependencies listed in environment.yaml?  The computer has a very recent driver and CUDA 12.2.

RaulPPelaez:
Trying to get to the bottom of this here: https://github.com/conda-forge/cuda-version-feedstock/issues/14

docs/source/installation.rst
sef43, line 1: the rst markup formatting is not quite correct on this file. At least looking at the version rendered by github: https://github.com/RaulPPelaez/torchmd-net/blob/docs_update/docs/source/installation.rst

README.md
sef43, line 1: can we just link to the installation.rst file instead of duplicating error prone instructions?

sef43 (changes requested):

README.md
RaulPPelaez: I agree!

RaulPPelaez:
I have not gotten a response yet from mamba guys. Perhaps the way to go here is not recommend mamba for the "from source" installation instructions.

sef43:
Alternatively we just point people to the CUDA docs and tell them to install nvcc + extra from the official NVIDIA channels?
https://docs.nvidia.com/cuda/cuda-installation-guide-linux/index.html#conda-installation
Making a full nvcc compiler setup from conda is good for CI and deployment but not ideal for a user who probably already has a working nvcc on their system

RaulPPelaez:
Well, official instructions include installing using conda from the nvidia channel. The other ones normally require root access, which I do not like given how we manage to get all the rest of the dependencies without it.
It is true though that in the end the target machine requires the correct driver anyway, so might as well get CUDA in the process. Let me give it a try and see if I can put everything in there...

sef43:
The latest conda install commands work for me for cuda 12

RaulPPelaez:
I am going to merge this.

peastman:
I'm trying to follow the instructions in this PR.  Conda has been running for about an hour trying to solve the environment for the `conda env create` command. :(

sef43:
Note that new and up to date conda will actually use libmamba solver backend. So for me the conda install was just as fast as the mamba one. But then I don't know if the original mamba bug in this issue still occurs.

https://conda.org/blog/2023-07-05-conda-libmamba-solver-rollout/

peastman:
So it might be faster but have the same problem as mamba...  How useful!

It did finally succeed, and it installed CUDA 12.0.  It's now working on the command to install the rest of the CUDA libraries.

peastman:
Everything eventually succeeded and I now have a working install.  It just took patience.  Thanks!

PR #252, opened 2024-01-18
Ani-2x dataloader

stefdoerr:
- Added a data loader for ANI-2x
- Fixed a bug with variable names for element energies (can you confirm this was wrong?)
- Fixed a bug where training would crash if log folder was missing


torchmdnet/datasets/ani.py
```diff
@@ -206,12 +208,12 @@ def get(self, idx):
 class ANI1(ANIBase):
     __doc__ = ANIBase.__doc__
     # Avoid sphinx from documenting this
-    ELEMENT_ENERGIES = {
```
stefdoerr, line 209: In line 49 there is this code `energy = sum(self._ELEMENT_ENERGIES[z] for z in atomic_numbers)`
which seems to assume that the classes have a `_ELEMENT_ENERGIES` variable. However here it's missing the underscore. Am I right in changing this?

torchmdnet/scripts/train.py
```diff
@@ -130,6 +131,7 @@ def get_args():
     if args.inference_batch_size is None:
         args.inference_batch_size = args.batch_size
 
+    os.makedirs(os.path.abspath(args.log_dir), exist_ok=True)
```
stefdoerr, line 134: Creates the log folder instead of getting an error in save_argparse

RaulPPelaez: Damn thanks for this. I was really tired of having to create the folder.

torchmdnet/datasets/ani.py
RaulPPelaez: Definetely a bug. Thanks!

RaulPPelaez (approved):

RaulPPelaez:
Thanks @stefdoerr ! ready to merge?

torchmdnet/scripts/train.py
guillemsimeon: 100%

stefdoerr:
Yeah fine by me

PR #260, opened 2024-01-25
Adding vector feature cutoff to ET

### LINKED ISSUE #254: Discontinuity in the ET potential
We found that there is a discontinuity in the torchmd-net ET potential when atoms cross the cutoff distance. I assume it is not the intended behavior of the potential. We noticed that in larger calculations with our own models, but it can easily be demonstrated on a system with two atoms using a model from the SPICE-models repository (https://github.com/openmm/spice-models). Therefore, the issue is not specific to the models we are developing.

The following code (using torchmd-net 0.8.1 installed with conda, the results are the same when I upgraded to 0.11.0) reproduces the issue, there is a discontinuity of about 3.6 kJ/mol at the distance of 10A, which is the cutoff used in this model. In our models with cutoff 5 A, similar step occurs at that distance. In the intermolecular potential of larger molecules, multiple steps can be found where the individual atoms cross the cutoff.

The code produces a warning at the cutoff distance saying that gradients are not calculated, but that is not the problem - it is the change in energy.

```
import sys
import torch
from torchmdnet.models.model import load_model

# Load the model
# As an example, model from the spice-models repository is used:
# https://github.com/openmm/spice-models/blob/main/five-et/model1.ckpt
model = load_model("model1.ckpt", derivative=True)

# Types - two atoms of selected elements, H and O
types = torch.tensor([10, 21], dtype=torch.long)

# Iterate distances from 9.9 to 10.1 A
dist = 9.9
step = 0.01
while dist < 10.1:
    # Construct the geometry
    pos = torch.tensor([[0,0,0],[dist,0,0]], dtype=torch.float32)
    # Compute energy and forces
    energy, forces = model.forward(types, pos)
    # Print distance and energy
    print("dist = %.3f   energy = %.5f" % (dist, energy))
    dist += step
```

RaulPPelaez:
The user in #254 discovered a discontinuity in the energy at the cutoff when using ET.
We found that this is caused by a quirk in the current architecture that cannot be avoided without invalidating previous checkpoints.

The message passing step in the current architecture, which is consistent with eq. 9 in [1], weights only scalar features with the cutoff function, not vector ones. Given that both distance projections (eq. 5) have a bias, this allows the arch to learn a non-vanishing offset on the energy.

We can solve this by modifying the message passing so that the cutoff also affects vector features. This amounts to taking away the $\phi(d_{ij})$ from eq. 8 and multiplying it instead at eq.9a, so that $s^1_{ij}, s^2_{ij}, s^3_{ij} = \text{split}(V_j\odot D^V_{ij}) \phi(d_{ij})$.

Since this invalidates checkpoints I am gating this change behind a new option, `vector_cutoff`, that defaults to false and if turned on implements this new behavior.

Enabling this option seems to slightly improve benchmarks, for instance, this is ET-MD17.yaml with and without (reference) vector_cutoff. Note that I did not do any kind of hparam optimization.
image

Moreover, since we are moving away from ET in general in favor of TensorNet, I am using this PR to mark ET as deprecated.

[1] https://arxiv.org/pdf/2202.02541.pdf

Closes #254

### DIFF
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -36,6 +36,9 @@ def create_model(args, prior_model=None, mean=None, std=None):
         args["check_errors"] = True
     if "static_shapes" not in args:
         args["static_shapes"] = False
+    if "et_vector_cutoff" not in args:
+        args["et_vector_cutoff"] = False
+
     shared_args = dict(
         hidden_channels=args["embedding_dimension"],
         num_layers=args["num_layers"],
@@ -85,6 +88,7 @@ def create_model(args, prior_model=None, mean=None, std=None):
             num_heads=args["num_heads"],
             distance_influence=args["distance_influence"],
             neighbor_embedding=args["neighbor_embedding"],
+            vector_cutoff=args["et_vector_cutoff"],
             **shared_args,
         )
     elif args["model"] == "tensornet":
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -36,8 +36,8 @@ def create_model(args, prior_model=None, mean=None, std=None):
         args["check_errors"] = True
     if "static_shapes" not in args:
         args["static_shapes"] = False
-    if "et_vector_cutoff" not in args:
-        args["et_vector_cutoff"] = False
+    if "vector_cutoff" not in args:
+        args["vector_cutoff"] = False
 
     shared_args = dict(
         hidden_channels=args["embedding_dimension"],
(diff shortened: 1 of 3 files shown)


RaulPPelaez:
cc @Honza-R

RaulPPelaez:
After careful consideration we decided not to deprecate ET for the time being.

RaulPPelaez:
@guillemsimeon please review

guillemsimeon (approved):
Looks good to me. Now both scalar and vector features will be weighted with the cutoff.

Honza-R:
Thanks for fixing the issue. We have very nice results obtained with ET, and discovered this at the very last moment. We're going to try TensorNet on the same problem once this works.

You mentioned that this invalidates the previous checkpoints - does this mean that they can't be used at all, or would it only change the results? I'd like to restart the training from what we already have, it took us some work to get the best models.

RaulPPelaez:
You should have no problems with old checkpoints. I left the old behavior as the default as to not invalidate checkpoints.
For new trainings you should definitely enable the vector_cutoff function. 
I would not expect an old model loaded with vector_cutoff=true (the fix) to behave well, but I have not tried. At least some epochs will be needed to readjust.

PR #261, opened 2024-01-26
Remove some old TODOs that are now solved

RaulPPelaez:


RaulPPelaez:
@stefdoerr could you take a look? You have been doing a lot of Datasets these days.

stefdoerr (approved):
makes sense since the PR is merged and if it works it works :)

PR #262, opened 2024-01-26
Unique COMP6 GDB molecule IDs

stefdoerr:
Tiny fix.

The molecule ids were not unique without the name of the file (they are repeated in different files of the same dataset) and it was causing issues in the evaluation when calculating relative conformer energies.


RaulPPelaez (approved):

RaulPPelaez:
Thanks!

PR #263, opened 2024-01-26
Update Examples

RaulPPelaez:
This PR is for updating the hparams in the examples, adding new ones and updating the trained checkpoints we provide for them.

### DIFF
--- a/examples/README.md
+++ b/examples/README.md
@@ -1,18 +1,20 @@
 # Examples
 
 ## Training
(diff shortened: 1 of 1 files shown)


RaulPPelaez:
Some tests are failing. Two of them are expected, since they compare the result of forward on the models with some stored reference (which I will update before merging).
However, one of them runs gradcheck and its failing for ET with vector_cutoff = true.

gradcheck compares the analytical (autograd) and numerical gradients, so it should not ever fail. This is worrying and I do not know why it fails.

guillemsimeon:
I have been looking at it for a while and I cannot find anything. At first I thought something about self loops, but everything seems fine. I will take a closer look again on Monday.

guillemsimeon (approved):
Done, right Raul?

guillemsimeon:
Nope, that’s it.

PR #264, opened 2024-01-26
Fix for COMP6v1 forces

stefdoerr:
the comp6 dataset has gradients labeled as forces by mistake


RaulPPelaez:
This makes a test fail, which expects the negative. I wonder why there is such an specific test there

guillemsimeon (approved):

PR #268, opened 2024-01-31
Add some type check in create_model

### LINKED ISSUE #267: Exception reloading model
When trying to reload a saved model with `load_model()`, it's failing with the error

```
get_neighbor_pairs_kernel(str strategy, Tensor positions, Tensor batch, Tensor box_vectors, bool use_periodic, float cutoff_lower, float cutoff_upper, int max_num_pairs, bool loop, bool include_transpose) -> ((Tensor, Tensor, Tensor, Tensor)):
Expected a value of type 'float' for argument 'cutoff_lower' but instead found type 'int'.
:
  File "/home/peastman/workspace/torchmd-net/torchmdnet/models/utils.py", line 256
        if batch is None:
            batch = torch.zeros(pos.shape[0], dtype=torch.long, device=pos.device)
        edge_index, edge_vec, edge_weight, num_pairs = get_neighbor_pairs_kernel(
                                                       ~~~~~~~~~~~~~~~~~~~~~~~~~ <--- HERE
            strategy=self.strategy,
            positions=pos,
```

At line 261 of utils.py, if I change

```python
            cutoff_lower=self.cutoff_lower,
            cutoff_upper=self.cutoff_upper,
```

to

```python
            cutoff_lower=float(self.cutoff_lower),
            cutoff_upper=float(self.cutoff_upper),
```

the error goes away.  I'm not sure whether that's the best place to fix it, but it seems a type conversion is needed somewhere.

RaulPPelaez:
When load_model or create_model are called explicitly there is no mechanism to enforce types of the arguments like it happens when train.py is used (via arg parse).

This means that something like:
```python
model = load_model(..., cutoff_upper=5)
```
fails because cutoff_upper should be a float but it is passed down as an integer.

This PR casts to float when necessary to avoid this.

Closes #267

### DIFF
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -46,10 +46,10 @@ def create_model(args, prior_model=None, mean=None, std=None):
         rbf_type=args["rbf_type"],
         trainable_rbf=args["trainable_rbf"],
         activation=args["activation"],
-        cutoff_lower=args["cutoff_lower"],
-        cutoff_upper=args["cutoff_upper"],
+        cutoff_lower=float(args["cutoff_lower"]),
+        cutoff_upper=float(args["cutoff_upper"]),
         max_z=args["max_z"],
-        check_errors=args["check_errors"],
+        check_errors=bool(args["check_errors"]),
         max_num_neighbors=args["max_num_neighbors"],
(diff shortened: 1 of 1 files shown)


RaulPPelaez:
@guillemsimeon review pls

guillemsimeon (approved):
awesome

PR #269, opened 2024-01-31
Fix inifinite recursion error on OSX

RaulPPelaez:
Circumvent a possible infinite recursion error in FloatCastDatasetWrapper in which getattr calls itself.
This problem only arose in OSX when a DataLoader uses more than one thread. In that case the dataset is pickled and causes an error.

This PR fixes it by making sure the getattr method in the aforementioned class does not call itself.


RaulPPelaez:
cc @sef43 can you please review?

sef43 (approved):
I no longer have access to a Mac to test manually. The build passes in https://github.com/conda-forge/torchmd-net-feedstock/pull/14 so looks good to me. Nice one!

PR #272, opened 2024-02-06
Add OpenMM integration example

RaulPPelaez:
Add an example script on how to use OpenMM-Torch to run MD. Also comment on MD engine integration in the docs


guillemsimeon (approved):
aprobao

PR #274, opened 2024-02-12
Add dtype parameter to External

RaulPPelaez:
This allows to use External with any dtype.
In addition to being useful for TorchMD integration, I normally use External to put models into CUDA graphs, as it handles all the nuisances transparently.

### DIFF
--- a/tests/test_calculator.py
+++ b/tests/test_calculator.py
@@ -12,49 +12,45 @@ from torchmdnet.models.model import load_model, create_model
 from utils import create_example_batch
 
 
-def test_compare_forward():
-    checkpoint = join(dirname(dirname(__file__)), "tests", "example.ckpt")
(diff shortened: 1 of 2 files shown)


RaulPPelaez:
@AntonioMirarchi do you see any issues with this? It should be retrocompatible.

RaulPPelaez:
With OpenMM one should not use External. TorchForce and TorchMD_Net get along well.

I made it so that dtype can be passed as a string. So in a yaml you can put:
```yaml
External:
   dtype: float32
```
Or any valid torch.dtype (double, float64, half,...)

AntonioMirarchi:
Niceee, I will try to run something to be sure but seems fine!

AntonioMirarchi:
LGTM, I can run sims with it using torchmd

PR #278, opened 2024-02-13
Mark standardize as deprecated

### LINKED ISSUE #277: Revisiting standardize
When enabled, the standardize functionality processes the whole Dataset and stores the mean and std of the energies (well, the "y" field) in it:

https://github.com/torchmd/torchmd-net/blob/6d8e3159cfb8bb971ecf7a2abd589735d79a7e53/torchmdnet/data.py#L172-L202

These are then used during prediction:

https://github.com/torchmd/torchmd-net/blob/6d8e3159cfb8bb971ecf7a2abd589735d79a7e53/torchmdnet/models/model.py#L358-L375

There are some issues we should consider with its current state:

1. There is no way to get unit std without shifting energies.
2. It has a potentially unwanted interaction with atomref.
3. For unit std it might be better to do batch normalization (which does not require going over the whole dataset before training)

I am opening this issue to start a discussion on what to do about these.

RaulPPelaez:
Following #277 and some internal discussion we are moving away from Dataset standardisation.
We believe its current implementation is not the best fit for NNPs but for the moment we are leaving the option there for backwards compatibility. It will be removed at some point.

We could instead simply remove the feature altogether or make it more sensible, like offering alternative approaches as to how the mean is computed.

This is your last chance if you have an opinion about this!
Closes #277

### DIFF
--- a/torchmdnet/data.py
+++ b/torchmdnet/data.py
@@ -14,7 +14,7 @@ from torch_geometric.data import Dataset
 from torchmdnet.utils import make_splits, MissingEnergyException
 from torchmdnet.models.utils import scatter
 from torchmdnet.models.utils import dtype_mapping
-
+import warnings
 
 class FloatCastDatasetWrapper(Dataset):
     """A wrapper around a torch_geometric dataset that casts all floating point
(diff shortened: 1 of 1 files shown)


RaulPPelaez:
Time is up! I am merging this.

PR #280, opened 2024-02-13
Better handling of prior configuration

RaulPPelaez:
I realized there is no way to set prior arguments from the yaml file. Also, the model can handle more than one prior, but there was no way to set more than one from the yaml file.
I updated the documentation to reflect this, which was wrong about it.

With this PR one can do something like:

```yaml
prior_model:
  Atomref:
    max_z: 100
  Coulomb:
      alpha: 1
      max_num_neighbors: 10
      distance_scale: 1
```

### DIFF
--- a/docs/source/priors.rst
+++ b/docs/source/priors.rst
@@ -26,10 +26,19 @@ For example, to use the Atomref prior via YAML, your configuration might look li
 .. code:: yaml
 
 	  prior_model: Atomref
-	  prior_args: 
-	     max_z: 100  # Optional argument for Atomref
+
+It is possible to configure more than one prior in this way:
+	     
+.. code:: yaml
+
+	  prior_model:
+	     Atomref: {} # No additional arguments
+	     Coulomb:
+	          alpha: 1
+		  max_num_neighbors: 10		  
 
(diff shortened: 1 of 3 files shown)


RaulPPelaez:
@stefdoerr please review, this will come in handy for #279 too.

RaulPPelaez:
I realized there is a much simpler change that can be done to instead allow this:
```yaml
prior_model:
  Atomref:
    max_z: 100
  Coulomb:
      alpha: 1
      max_num_neighbors: 10
      distance_scale: 1
```

peastman:
We already support setting args for multiple priors with exactly this syntax.  It was added in #134.  See in particular https://github.com/torchmd/torchmd-net/pull/134#issuecomment-1302469022.

RaulPPelaez:
Thanks for the ref!. It was undocumented and the argparser was preventing it from being used from the CLI.

stefdoerr:
By the way I would not bother too much with CLI arguments. I think passing a yaml file is the more usual use case, parsing CLI is a pain

stefdoerr (approved):

PR #281, opened 2024-02-14
Clean up some datasets

RaulPPelaez:
I found some errors and a lot of repeated code in the ANI datasets.

This PR fixes it.

### DIFF
--- a/torchmdnet/datasets/ani.py
+++ b/torchmdnet/datasets/ani.py
@@ -45,13 +45,15 @@ class ANIBase(MemmappedDataset):
     def raw_file_names(self):
         raise NotImplementedError
 
-    def compute_reference_energy(self, atomic_numbers):
-        atomic_numbers = np.array(atomic_numbers)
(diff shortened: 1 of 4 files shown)


torchmdnet/datasets/ani.py
```diff
@@ -134,18 +149,16 @@ def sample_iter(self, mol_ids=False):
...
-        refs[6] = -37.8302333826 * self.HARTREE_TO_EV  # C
-        refs[7] = -54.5680045287 * self.HARTREE_TO_EV  # N
-        refs[8] = -75.0362229210 * self.HARTREE_TO_EV  # O
-
-        return refs.view(-1, 1)
-
 
 class ANI1XBase(ANIBase):
+    __doc__ = ANIBase.__doc__
+    _ELEMENT_ENERGIES = {
```
stefdoerr, line 155: I think moving this to the base class might be wrong actually. ANI1CCX also is a child of this class but I'm relatively sure the reference energies of coupled cluster cannot match the ones of the DFT of ANI-1x

stefdoerr (approved):
lgtm

PR #285, opened 2024-02-16
Fix EnergyRefRemover

stefdoerr:
Fix the EnergyRefRemover which was accidentally removing the total sum of all atomref energies of the whole batch from the energies


RaulPPelaez (approved):
Perfect, thanks Stefan.

PR #287, opened 2024-02-20
Use switching function for Coulomb prior

peastman:
Originally I scaled the Coulomb prior by $\text{erfc}(\alpha r)$ to reduce its effect at short distances.  It turned out this didn't work well because it was still too short ranged.  Even at distances of only 1-2 A there was already a large component of Coulomb energy.  I changed it to a cosine switching function which works much better.

### DIFF
--- a/docs/source/priors.rst
+++ b/docs/source/priors.rst
@@ -31,11 +31,12 @@ It is possible to configure more than one prior in this way:
 	     
 .. code:: yaml
 
-	  prior_model:
-	     Atomref: {} # No additional arguments
-	     Coulomb:
-	          alpha: 1
-		  max_num_neighbors: 10		  
+    prior_model:
+        Atomref: {} # No additional arguments
+        Coulomb:
+            lower_switch_distance: 4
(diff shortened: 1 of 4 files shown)


peastman:
I think the test failure was caused by #286.  It's also failing on the main branch.

stefdoerr:
Pull from the main branch, that crash is fixed.

stefdoerr:
Fixed in main now

RaulPPelaez:
Looks good to me. This invalidates checkpoints using the Coulomb prior. I get that it is for the better but it is starting to happen a lot.
We really need to cook versioning into the checkpoints...

RaulPPelaez:
I added the versioning functionality in #288

peastman:
I don't think anyone has ever used the Coulomb prior.  It just didn't work very well.  I wouldn't worry about breaking compatibility with it.

peastman:
Tests are passing now.

RaulPPelaez (approved):
LGTM Thanks Peter

PR #289, opened 2024-02-21
Allow appending extra values to embedding vector

peastman:
This feature lets you inject extra per-sample or per-atom values into the model.  They are appended to the embedding vector for each atom, then a linear layer mixes them in and reduces the vector back down to its original size.  In the config file, you specify one or more extra fields from the dataset that should be added to the embedding:

```yaml
extra_embedding: partial_charges
```

or

```yaml
extra_embedding:
  - partial_charges
  - total_charge
```

Each one can have a single value per sample (for a global scalar like total charge) or a value for each atom.

When calling the model for inference, you specify them with `extra_args`:

```python
model(z, pos, extra_args={'partial_charges':partial_charges)
```

See https://github.com/torchmd/torchmd-net/issues/26#issuecomment-1939258460 for discussion.

### DIFF
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -38,6 +38,12 @@ def create_model(args, prior_model=None, mean=None, std=None):
         args["static_shapes"] = False
     if "vector_cutoff" not in args:
         args["vector_cutoff"] = False
+    if "extra_embedding" not in args:
+        extra_embedding = None
+    elif isinstance(args["extra_embedding"], str):
+        extra_embedding = [args["extra_embedding"]]
+    else:
+        extra_embedding = args["extra_embedding"]
 
     shared_args = dict(
         hidden_channels=args["embedding_dimension"],
@@ -57,6 +63,7 @@ def create_model(args, prior_model=None, mean=None, std=None):
             else None
         ),
         dtype=dtype,
+        extra_embedding=extra_embedding
     )
 
     # representation network
@@ -370,7 +377,7 @@ class TorchMD_Net(nn.Module):
             If this is omitted, periodic boundary conditions are not applied.
             q (Tensor, optional): Atomic charges in the molecule. Shape: (N,).
             s (Tensor, optional): Atomic spins in the molecule. Shape: (N,).
-            extra_args (Dict[str, Tensor], optional): Extra arguments to pass to the prior model.
+            extra_args (Dict[str, Tensor], optional): Extra arguments to pass to the model.
 
         Returns:
             Tuple[Tensor, Optional[Tensor]]: The output of the model and the derivative of the output with respect to the positions if derivative is True, None otherwise.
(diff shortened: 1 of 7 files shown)


torchmdnet/models/model.py
```diff
@@ -370,7 +377,7 @@ def forward(
             If this is omitted, periodic boundary conditions are not applied.
             q (Tensor, optional): Atomic charges in the molecule. Shape: (N,).
             s (Tensor, optional): Atomic spins in the molecule. Shape: (N,).
-            extra_args (Dict[str, Tensor], optional): Extra arguments to pass to the prior model.
+            extra_args (Dict[str, Tensor], optional): Extra arguments to pass to the model.
```
RaulPPelaez, line 380: I would say to the model AND the prior model.

RaulPPelaez:
The design and implementation are great and retrocompatible.
@guillemsimeon, could you comment on the architecture changes? 
Peter made it so that the extra arguments are mixed with the embedding using a Linear layer:
https://github.com/torchmd/torchmd-net/blob/13926ad59446ab1c8bdf0354845028d3d30cff8f/torchmdnet/models/tensornet.py#L308-L311
https://github.com/torchmd/torchmd-net/blob/13926ad59446ab1c8bdf0354845028d3d30cff8f/torchmdnet/models/tensornet.py#L345-L347

guillemsimeon:
We already discussed the changes in the issue.

torchmdnet/models/model.py
peastman: I was thinking of "the model" as encompassing all parts of the model (representation model, output model, and priors).

peastman:
This is finished and ready to merge as far as I'm concerned.  I'm getting good results with it.  Can someone review it?

giadefa:
we cannot merge this for now,  use it in a branch for now.
Once we finish what we are doing we will come back to it to see how it
performs and if it makes sense to merge it into main.

peastman:
Why can't we merge it?  What are you doing that conflicts with it?

If this isn't going to be run as an open source project, I can't use it.  I need a stable, reliable base to build my research on.  That means development is done in public, and I can add features that are essential to my research.  If you're going to reject changes without explanation because they conflict with things you're doing in secret and can't tell me about, then I need to end my involvement in this project.

giadefa:
I don't think that what you are saying is right at many levels. Simply, as
explained before, the change that you propose has not yet been tested by
us. It is a model change, not a cosmetic one. As I wrote before, it might
be superseded by another change that we are working on and that we have
tested already. Use it in your branch for a bit until we can assess it. If
it is better, we will of course merge it, but we also test many things and
not all go to main. Happy to clarify further tomorrow.

peastman:
As I said before, it is *not* a change to the existing model.  It's a completely new feature.  If you don't use the new feature, nothing in your model changes the slightest bit.  It breaks nothing.  It conflicts with nothing.

I've already done extensive testing of it and found that it substantially improves accuracy.  I can post numbers if you want to see them. What additional validation do you need?

And we really need to do development in public.  Being told you are working on "something" that "might" supersede this in the future, but it's secret and you can't tell me about it, is very frustrating and not useful.  In an open source project, decisions get made in public based on public information.  You can have your own internal branch where you work on things you don't want to share.  But you can't block development of the public version for private reasons that you won't tell anyone.

If I'm going to base my research on this project, it needs to be a stable base to build on.  That means having a clear process for reviewing and merging PRs.  I have a whole series of features I want to add.  I can't create a branch with a new feature, then another branch with another feature based on top of that one, then a third branch with a third feature based on the second one, and so on for months, with my version of the code steadily diverging from the main repo and never knowing whether the very first PR will get rejected and break everything I did after.

PR #291, opened 2024-02-23
Added versioneer and updated setuptools building

stefdoerr:


pyproject.toml
```diff
@@ -0,0 +1,40 @@
...
+[tool.versioneer]
+VCS = "git"
+style = "pep440"
+versionfile_source = "torchmdnet/_version.py"
+versionfile_build = "torchmdnet/_version.py"
+tag_prefix = ""
+parentdir_prefix = "torchmdnet-"
+
+[build-system]
+requires = ["setuptools", "toml", "versioneer[toml]==0.28", "torch<2.2"]
```
RaulPPelaez, line 39: There are probably more dependecies, is it required that they are listed twice? (environment.yaml and here) I feel like this will get desynchronized rather quickly.
Also I do not think there is a restriction on torch version.

stefdoerr: These are the requirements for _building_ the package, not for running it. So practically everything that's required to run `setup.py`

stefdoerr: The restriction of torch version I took from the environment.yaml but no issues on my side with removing it. I doubt it affects the package building

stefdoerr: Yeah I am actually seeing that issue right now where it's trying to find CUDA but it's not installed in the venv. I guess we will have to ditch the venv approach. But we can use versioneer regardless. Let me fix it

RaulPPelaez: I got PTSD from when I was trying to port NNPOps to pip and learned a bit about pyproject.toml. In my experience the venv thing will just not ever work. It assumes everything you need is in pip AND that packages are compatible between versions of the dependencies. I often ran into the build env installing torch X but the run env getting torch Y.
 I cannot fathom what they were thinking when they designed it that way. I can only assume I am missing something fundamental about it -.-

stefdoerr: You can still use pyproject.toml without venvs though I think. You just need to specify the correct build engine but I'm not going to bother. This works right now so let's stick to that :P

stefdoerr:
Done I think. Tests pass, so we should be fine

docs/source/installation.rst
```diff
@@ -54,7 +54,7 @@ It is recommended to install the same version as the one used by torch.
 
 .. code-block:: shell
 
-   conda install -c conda-forge cuda-nvcc cuda-libraries-dev cuda-version "gxx<12" pytorch=*=*cuda*
+   conda install -c conda-forge cuda-nvcc cuda-libraries-dev cuda-version "gxx<12" "pytorch=*=*cuda*"
```
stefdoerr, line 57: also fixed this because it bothered me :)

RaulPPelaez:
Looking at the versioneer README it seems like one of the steps requires running "versioneer install --no-vendor". 
I understand this is what generated the big _version.py file and it is not required to run it again? (unless to update versioneer I guess).
Also, does this add versioneer as a dependency?

docs/source/installation.rst
RaulPPelaez: The " surrounds gxx because otherwise bash will interpret it as "my_command<12" (a.i. pipe the file "12" as input). I have never found it a problem to not add the quotes when using wildcards. But I can see it being a problem.

stefdoerr:
```sh
(torchmd-net2)  yyy@xxx  ~/Work/torchmd-net   versioning ±  mamba install -c conda-forge cuda-nvcc cuda-libraries-dev cuda-version "gxx<12" pytorch=*=*cuda*
zsh: no matches found: pytorch=*=*cuda*
```
might just be my `zsh` but it doesn't work for me

RaulPPelaez:
It also appears that versioneer is just a build time dependency, which is nice.

stefdoerr:
Yes, since we ship that big `_version.py` file you don't need to add it as a runtime dep

Issue #296, opened 2024-02-29
How to run custom training loops?


ljmartin:
Hi TorchMD-Net,
Following up from a twitter post of Guillem. 

I'm interested in running 'custom' training loops. As I understand it, currently the `torchmd-train` routine will learn a mapping between atoms and per-atom energies, and maybe molecules <-> per-molecule energies (i.e. summing atom energies into a per-molecule level). I'd like to try other things that don't fit into a PyTorch Lightning format - to think of some examples:
- using ranking loss functions over a batch, instead of a direct mapping to an energy
- learning on relative energies rather than absolute. Perhaps for sets of tautomers, where each molecule has 1 to 10 tautomers, and the absolute energy of each is not known or important, but the free energy difference between them is known. Similarly one could imagine learning relative binding free energies in a congeneric series 

Anyway, question is: is it suitable to just instantiate a model like TorchMD_ET, run a forward pass over it, and slot that into any custom training&loss loop? 
Thanks for the tool and congrats on the nice publication
lewis

RaulPPelaez:
Hi! Thanks for the kind words. Guillem is our executive public relationships officer indeed.

The principal "model" abstraction in TorchMD-Net is just a regular torch.nn.Module, which you can train as you would any other module:
https://torchmd-net.readthedocs.io/en/latest/models.html#available-models
The model can be created with this function:
https://torchmd-net.readthedocs.io/en/latest/generated/torchmdnet.models.model.html#torchmdnet.models.model.create_model
Where the parameters are any of the ones supported by torchmd-train:
https://torchmd-net.readthedocs.io/en/latest/torchmd-train.html#configuration-file

Your feature request (custom training loops), while well supported, could be better documented. So let me go ahead and add a section for it. I will link it to you here when its done.


Having said that, you should now that:
1. You are free to create a dataset providing anything other than energies/forces. Although we do not currently provide a way to configure custom loss functions it is something in our roadmap.
2. We do support training on relative energies, perhaps you can make use of it for your purposes https://torchmd-net.readthedocs.io/en/latest/models.html#training-on-relative-energies

RaulPPelaez:
I added a section on custom training loops to the docs, check it out! https://torchmd-net.readthedocs.io/en/latest/models.html

RaulPPelaez:
Depending on your usecase you might also just use the representation model (i.e TensorNet, Equivariant Transformer...) which will return (N_atoms, hidden_features) scalar and/or vector features.
[...]

PR #298, opened 2024-03-01
Add custom training loop to the docs

RaulPPelaez:
Add a more thorough description of the TorchMD_Net model and an example on how to carry out custom training loops.

I hanged the compiled version of this PR here for reviewing purposes https://torchmd-net.readthedocs.io/en/update_docs/models.html


RaulPPelaez:
@guillemsimeon, @AntonioMirarchi  pls review. The code examples run fine, but maybe I missed something.

guillemsimeon:
Looks good to me. I think in #296 Lewis is interested in just instantiating TorchMD_ET, not the whole TorchMD_Net thing. But for the docs, I find the description very illustrative. Once this is merged, we can refer to this PR and we can tell Lewis that either you design a custom output model to be integrated into the whole TorchMD_Net thing by using create_model, or you can instantiate any representation model, but not using create_model in that case.

guillemsimeon (approved):

PR #301, opened 2024-03-02
Added vector output from skew-symmetric tensor

### LINKED ISSUE #297: Vector output from TensorNet
##### Feature Request
Currently the vector output is set to `None`. This could be easily calculated based on skew symmetric matrix based on the paper. 
We can add the following function in order to do it,

```python
def skew_tensor_to_vector(tensor):
    """Extracts a vector from a skew-symmetric tensor.
    Equation (3) in the paper
    """
    return torch.stack((tensor[:, :, 1, 2], tensor[:, :, 2, 0], tensor[:, :, 0, 1]), dim=-1)
```

And in the line https://github.com/torchmd/torchmd-net/blob/fdd4dac8852f1c42906e7c7a5f4ffa70319a41b2/torchmdnet/models/tensornet.py#L271

do, 
```
return x, skew_tensor_to_vector(A), z, pos, batch
```

It might be useful if someone wants to use an `EquivariantScalar` or `EquivariantVectorOutput` modules on top of `TensorNet`.

shenoynikhil:
Allowing vector output from TensorNet. This will allow the use of `EquivariantScalar` or `EquivariantVectorOutput` from `TensorNet`.

Resolves #297.

### DIFF
--- a/torchmdnet/models/tensornet.py
+++ b/torchmdnet/models/tensornet.py
@@ -48,6 +48,10 @@ def vector_to_symtensor(vector):
     S = 0.5 * (tensor + tensor.transpose(-2, -1)) - I
     return S
 
+def skewtensor_to_vector(tensor):
+    '''Converts a skew-symmetric tensor to a vector.'''
+    return torch.stack((tensor[:, :, 1, 2], tensor[:, :, 2, 0], tensor[:, :, 0, 1]), dim=-1)
+
 
 def decompose_tensor(tensor):
     """Full tensor decomposition into irreducible components."""
@@ -265,10 +269,13 @@ class TensorNet(nn.Module):
         x = torch.cat((tensor_norm(I), tensor_norm(A), tensor_norm(S)), dim=-1)
         x = self.out_norm(x)
         x = self.act(self.linear((x)))
+        v = skewtensor_to_vector(A)
+        v = v.transpose(1, 2)
         # # Remove the extra atom
         if self.static_shapes:
             x = x[:-1]
-        return x, None, z, pos, batch
+            v = v[:-1]
+        return x, v, z, pos, batch
 
 
 class TensorEmbedding(nn.Module):
--- a/torchmdnet/models/tensornet.py
+++ b/torchmdnet/models/tensornet.py
@@ -124,6 +124,7 @@ class TensorNet(nn.Module):
             (default: :obj:`True`)
         check_errors (bool, optional): Whether to check for errors in the distance module.
             (default: :obj:`True`)
+        vector_output (bool, optional): Whether to return 
     """
 
     def __init__(
@@ -143,6 +144,7 @@ class TensorNet(nn.Module):
         check_errors=True,
         dtype=torch.float32,
         box_vecs=None,
+        vector_output=False
     ):
         super(TensorNet, self).__init__()
 
@@ -214,6 +216,7 @@ class TensorNet(nn.Module):
             box=box_vecs,
             long_edge_index=True,
         )
+        self.vector_output = vector_output
 
         self.reset_parameters()
 
@@ -269,12 +272,21 @@ class TensorNet(nn.Module):
         x = torch.cat((tensor_norm(I), tensor_norm(A), tensor_norm(S)), dim=-1)
         x = self.out_norm(x)
         x = self.act(self.linear((x)))
-        v = skewtensor_to_vector(A)
-        v = v.transpose(1, 2)
-        # # Remove the extra atom
+        # Remove the extra atom
         if self.static_shapes:
             x = x[:-1]
-            v = v[:-1]
+
+        # calculate vector_output if needed
+        v = None
+        if self.vector_output:
+            # (n_atoms, hidden_channels, 3, 3) -> (n_atoms, hidden_channels, 3)
+            v = skewtensor_to_vector(A)
+            # (n_atoms, hidden_channels, 3) -> (n_atoms, 3, hidden_channels)
+            v = v.transpose(1, 2)
+
+            if self.static_shapes:
+                v = v[:-1]
+
         return x, v, z, pos, batch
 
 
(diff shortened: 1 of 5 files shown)


guillemsimeon:
Thanks a lot! Can you add the equivariance test, please? I understood what you meant in the issue (it just confused me to see them denotes as 'forces' and not specifying further :) )

shenoynikhil:
I should be able to add it today. Sorry for the delay.

guillemsimeon:
Don’t rush! Any day will be fine.

Thank you!

torchmdnet/models/tensornet.py
```diff
@@ -265,10 +269,13 @@ def forward(
         x = torch.cat((tensor_norm(I), tensor_norm(A), tensor_norm(S)), dim=-1)
         x = self.out_norm(x)
         x = self.act(self.linear((x)))
+        v = skewtensor_to_vector(A)
+        v = v.transpose(1, 2)
```
RaulPPelaez, line 273: Would it not be possible to generate this tensor already transposed? By changing how the tensor is stacked.

torchmdnet/models/tensornet.py
```diff
@@ -265,10 +269,13 @@ def forward(
...
         # # Remove the extra atom
         if self.static_shapes:
             x = x[:-1]
-        return x, None, z, pos, batch
+            v = v[:-1]
+        return x, v, z, pos, batch
```
RaulPPelaez, line 278: I am worried about the consequences of returning a vector but not using it down the line. On one hand this can mess with autograd and on the other it is wasted computation when the vector features are not used by the output model.

torchmdnet/models/tensornet.py
guillemsimeon: I explicitly asked him to build skewtensor_to_vector acting and giving outputs in a way we use vector_to_skewtensor and vector_to_symtensor, which implies having to transpose later on, it is in the issue.

shenoynikhil: I remember checking what `TorchMD_ET` returns as v and its shaped like `(num_atoms, 3, hidden_channels)` whereas `A` is shaped like `(num_atoms, hidden_channels, 3, 3)`.

guillemsimeon: Yes, I imagined. Linears act on the last dimension (they mix hidden_channels), and since for the ET the default output is EquivariantScalar, we could already take for granted that v will be used by linears. I would say, leave it in the way you prefer, because maybe we will have to revisit with @RaulPPelaez again? I have just realized (lol) that if the do everything from the beginning (3,3,hidden) we might be able to remove those ugly permutes from the tensor linears.

RaulPPelaez: yes exactly. My only problem with this to get it merged is that the vector features should not be generated by default.
An option to TensorNet is not ideal, but I do not see another way out of it...

guillemsimeon: then let's add the flag: vector_output

torchmdnet/models/tensornet.py
shenoynikhil: Just added it.

torchmdnet/models/tensornet.py
shenoynikhil: Just updated it with `vector_output` flag.

shenoynikhil:
> Can you add the equivariance test, please?
Added an equivariance test.

guillemsimeon:
Hey! Sorry for the delay. Revisiting this:

Your approach seems correct @shenoynikhil, however I think I would not proceed by adding a new model that is called equivariant-tensornet. This makes me think, @RaulPPelaez, that we should get rid of the is_equivariant flag, and directly modify the places where the ET is used (yaml example files, for example), and directly include there EquivariantScalar, instead of Scalar, as output model, where it is used. What do you think? I don't know how many things this change would impact, but in any case I think it is worth doing it, the flag does not make much sense at this point...

For me, I would say it is fine if TensorNet defaults to returning x, v instead of x, None. Even though TensorNet uses Scalar (not EquivariantScalar), for energy predictions, I have the impression that nothing would happen. As I said above at some point, v comes from A, which has already been used by learnable parameters to generate x. Using Scalar, no new learnable things would act independently on v, and therefore we should not get an error pointing to not using parameters that are trainable in the forward.

What do you think guys?

torchmdnet/models/tensornet.py
```diff
@@ -120,6 +124,7 @@ class TensorNet(nn.Module):
             (default: :obj:`True`)
         check_errors (bool, optional): Whether to check for errors in the distance module.
             (default: :obj:`True`)
+        vector_output (bool, optional): Whether to return
```
RaulPPelaez, line 127: This sentence is cut off mid way it seems

RaulPPelaez:
> I would not proceed by adding a new model that is called equivariant-tensornet

I agree because it kind of implies that there is a non-equivariant tensornet.
I would be more comfortable with tensornet setting `is_equivariant` (and thus selecting the Equivariant* output models) if vector_output is set to true. This option could come from train.py as "tensornet-vector-output" maybe.
It is not standard for us to refer to different flavors of a model with different names, but I am not totally against it personally. In this particular case I do not think it is a good idea, imagine I have "vector-tensornet" on one hand and "tensornet-whatever" in another and I would like "vector-tensornet-whatever", there would be no way to get that.



> For me, I would say it is fine if TensorNet defaults to returning x, v instead of x, None. Even though TensorNet uses Scalar (not EquivariantScalar), for energy predictions, I have the impression that nothing would happen

I do not like this because even if autograd eats it it is extra unnecessary work. I believe gating the functionality behind an user-provided function or a model name is best in this case. 
 
> we should get rid of the is_equivariant flag, and directly modify the places where the ET is used (yaml example files, for example), and directly include there EquivariantScalar, instead of Scalar, as output model, where it is used.

I agree with this, but I believe it is out of the scope of this PR. It would be great if you open an issue so we have it in our list...
For retrocompatibility create_model can simply compare "model_name" with "equivariant-transformer" to decide whether to transform Scalar into EquivariantScalar.

PR #304, opened 2024-03-05
pq

guillemsimeon:
Add small hack to be able to test partial charges with SPICE. This can be the base PR to come up with a solution to total charge, partial charges, spins, etc.

### DIFF
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -379,7 +379,7 @@ class TorchMD_Net(nn.Module):
         batch = torch.zeros_like(z) if batch is None else batch
 
         # trick to incorporate SPICE pqs
-        # set charge: true in yaml
+        # set charge: true in yaml ((?) currently I do it)
         q = extra_args["pq"]
         
         if self.derivative:
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -378,6 +378,10 @@ class TorchMD_Net(nn.Module):
         assert z.dim() == 1 and z.dtype == torch.long
         batch = torch.zeros_like(z) if batch is None else batch
 
+        # trick to incorporate SPICE pqs
+        # set charge: true in yaml
+        q = extra_args["pq"]
+        
         if self.derivative:
             pos.requires_grad_(True)
         # run the potentially wrapped representation model
--- a/torchmdnet/models/tensornet.py
+++ b/torchmdnet/models/tensornet.py
@@ -239,7 +239,8 @@ class TensorNet(nn.Module):
         # Total charge q is a molecule-wise property. We transform it into an atom-wise property, with all atoms belonging to the same molecule being assigned the same charge q
         if q is None:
             q = torch.zeros_like(z, device=z.device, dtype=z.dtype)
-        else:
+        # if not atom-wise, make atom-wise (pq is already atom-wise)
+        if z.shape != q.shape:
             q = q[batch]
         zp = z
         if self.static_shapes:


torchmdnet/models/model.py
```diff
@@ -378,6 +378,10 @@ def forward(
         assert z.dim() == 1 and z.dtype == torch.long
         batch = torch.zeros_like(z) if batch is None else batch
 
+        # trick to incorporate SPICE pqs
+        # set charge: true in yaml ((?) currently I do it)
+        q = extra_args["pq"]
```
RaulPPelaez, line 383: I would move instead to something like removing "q" and "s" from the forward call of the representation models and instead send everything as extra_args, or atomic_labels:
```python
        atomic_labels = {}
        if q is not None:
             atomic_labels["total_charges"] = q
        if s is not None:
             atomic_labels["spin"] = s 
        # run the potentially wrapped representation model
        x, v, z, pos, batch = self.representation_model(
            z, pos, batch, box=box, atomic_labels=atomic_labels, extra_args=extra_args
        )
```

RaulPPelaez:
You could avoid creating a tensor of zeros with the charges here:
https://github.com/torchmd/torchmd-net/blob/fae79bd9e8e96ab235c52d5e54e33c3be9e4d05d/torchmdnet/models/tensornet.py#L240-L243
https://github.com/torchmd/torchmd-net/blob/fae79bd9e8e96ab235c52d5e54e33c3be9e4d05d/torchmdnet/models/tensornet.py#L496
If you set "q" (please lets change the name of that variable) to something like this:
```python
        if q is None:
            q = 0
        else:
            q = q[batch, None, None, None]
```

torchmdnet/models/model.py
guillemsimeon: it seems nice

peastman:
I previously tested this method for injecting partial charges and found it didn't work at all.  The loss got much worse.  In contrast, the method in #289 does work and makes the loss better.

peastman:
Yes, we did already have this discussion. :)  My point is that this shouldn't be merged until we do a full evaluation of both PRs and determine which one works better.

Also, #289 truly is a generic method that supports arbitrary sets of global and per-atom properties.  This one is hardcoded to just a single property which must be called `pq`, and it isn't clear how it could be generalized to support multiple properties.

guillemsimeon:
Sorry Peter, perhaps I was not clear enough: I meant agnostic in the sense
that we are trying to come up with a unified way to feed the models with
all these extra attributes, rename them to have clearer names, being
compatible with priors, etc, regardless of how they are manipulated inside
the models. Regarding this way of dealing with them, I have tests with
total charges, partial charges, and spins, and results are pretty
convincing (in our setting at least, not using any prior, I cannot specify
further since I am writing a preprint, but I will be happy to discuss
privately if you want to reproduce it) But yes, this is not going to be
merged at the moment.

AntonioMirarchi:
What do you think about this setup in the yaml file:
```
model_extra_args:
  total_charge: 
    learnable: true
    coeff: 0.1
  partial_charge: 
    learnable: true
    coeff: 0.1
  spin: 
    learnable: false
    coeff: 1
```
The keys of the dict will be used to define which args of `extra_args` should go to the model so the prior behavior is preserved as it is now. Then the coeff is used as starting parameter for the specific arg, if learnable `requires_grad` is set to true. In the init  a `nn.ParameterList` is initialized with length equal to the  `model_extra_args.keys()`.

RaulPPelaez:
You are assuming the extra_args are going to just be multiplied by a constant (as guillem does now), this might not be the case in the future, when perhaps we lean towards Peter's strategy of embeddings, or we use some kind of small MLP to compute the prefactors.
OTOH I do not really like the name "model_extra_args". Why would "total_charge" be extra but "box" is just an optional argument? I do not have a better alternative for you, just saying...

PR #306, opened 2024-03-15
[WIP] Introduce additional_labels

AntonioMirarchi:
This PR proposes several enhancements to enable the usage of extra arguments in both training and inference. The key modifications include
- [x] standardizing the naming conventions for all extra fields across different data loaders;
- [x] removing q and s as explicit args in the forward;
- [x] introducing additional labels through a new dictionary for model initialization, example:
```
additional_labels:
     tensornetq:
        label: total_charge
        learnable: true
        init_value: 0.1
```
- [x] add TensornetQ nn.Module
- [x] extra args expansion from molecule-wise to atom-wise inside the model
- [x] tensornet_q tested with total_charge and partial_charges, reproducibility achieved   
- [x] rewrite some tests according to the new methods
At the moment only 'tensornet_q' supported as additional_methods but more can be added

### DIFF
--- a/torchmdnet/datasets/ace.py
+++ b/torchmdnet/datasets/ace.py
@@ -273,7 +273,7 @@ class Ace(MemmappedDataset):
 
                     # Create a sample
                     args = dict(
-                        z=z, pos=pos, y=y.view(1, 1), neg_dy=neg_dy, q=q, pq=pq, dp=dp
+                        z=z, pos=pos, y=y.view(1, 1), neg_dy=neg_dy, total_charge=q, partial_charges=pq, dipole_moment=dp
                     )
                     if mol_ids:
                         args["mol_id"] = mol_id
--- a/torchmdnet/datasets/ace.py
+++ b/torchmdnet/datasets/ace.py
@@ -47,7 +47,7 @@ class Ace(MemmappedDataset):
     - `positions`: Atomic positions. Units: Angstrom.
     - `forces`: Forces on the atoms. Units: eV/Å.
     - `partial_charges`: Atomic partial charges. Units: electron charges.
-    - `dipole_moment` (version 1.0) or `dipole_moments` (version 2.0): Dipole moment (a vector of three components). Units: e*Å.
+    - `dipole_moment` (version 1.0) or `dipole_moment` (version 2.0): Dipole moment (a vector of three components). Units: e*Å.
     - `formation_energy` (version 1.0) or `formation_energies` (version 2.0): Formation energy. Units: eV.
 
     Each dataset should also have an `units` attribute specifying its units (i.e., `Å`, `eV`, `e*Å`).
@@ -146,7 +146,7 @@ class Ace(MemmappedDataset):
             transform,
             pre_transform,
             pre_filter,
-            properties=("y", "neg_dy", "q", "pq", "dp"),
+            properties=("y", "neg_dy", "total_charge", "partial_charges", "dipole_moment"),
         )
 
     @property
@@ -188,14 +188,14 @@ class Ace(MemmappedDataset):
             assert neg_dy.shape == pos.shape
 
             assert conf["partial_charges"].attrs["units"] == "e"
-            pq = pt.tensor(conf["partial_charges"][:], dtype=pt.float32)
-            assert pq.shape == (n_atoms,)
+            partial_charges = pt.tensor(conf["partial_charges"][:], dtype=pt.float32)
+            assert partial_charges.shape == (n_atoms,)
 
             assert conf["dipole_moment"].attrs["units"] == "e*Å"
-            dp = pt.tensor(conf["dipole_moment"][:], dtype=pt.float32)
-            assert dp.shape == (3,)
+            dipole_moment = pt.tensor(conf["dipole_moment"][:], dtype=pt.float32)
+            assert dipole_moment.shape == (3,)
 
-            yield pos, y, neg_dy, pq, dp
+            yield pos, y, neg_dy, partial_charges, dipole_moment
 
     @staticmethod
     def _load_confs_2_0(mol, n_atoms):
@@ -213,19 +213,19 @@ class Ace(MemmappedDataset):
         assert all_neg_dy.shape == all_pos.shape
 
         assert mol["partial_charges"].attrs["units"] == "e"
-        all_pq = pt.tensor(mol["partial_charges"][...], dtype=pt.float32)
-        assert all_pq.shape == (n_confs, n_atoms)
+        all_partial_charges = pt.tensor(mol["partial_charges"][...], dtype=pt.float32)
+        assert all_partial_charges.shape == (n_confs, n_atoms)
 
         assert mol["dipole_moments"].attrs["units"] == "e*Å"
-        all_dp = pt.tensor(mol["dipole_moments"][...], dtype=pt.float32)
-        assert all_dp.shape == (n_confs, 3)
+        all_dipole_moment = pt.tensor(mol["dipole_moments"][...], dtype=pt.float32)
+        assert all_dipole_moment.shape == (n_confs, 3)
 
-        for pos, y, neg_dy, pq, dp in zip(all_pos, all_y, all_neg_dy, all_pq, all_dp):
+        for pos, y, neg_dy, partial_charges, dipole_moment in zip(all_pos, all_y, all_neg_dy, all_partial_charges, all_dipole_moment):
             # Skip failed calculations
             if y.isnan():
                 continue
 
-            yield pos, y, neg_dy, pq, dp
+            yield pos, y, neg_dy, partial_charges, dipole_moment
 
     def sample_iter(self, mol_ids=False):
         assert self.subsample_molecules > 0
@@ -261,9 +261,9 @@ class Ace(MemmappedDataset):
 
                 z = pt.tensor(mol["atomic_numbers"], dtype=pt.long)
                 fq = pt.tensor(mol["formal_charges"], dtype=pt.long)
-                q = fq.sum()
+                total_charge = fq.sum()
 
-                for i_conf, (pos, y, neg_dy, pq, dp) in enumerate(
+                for i_conf, (pos, y, neg_dy, partial_charges, dipole_moment) in enumerate(
                     load_confs(mol, n_atoms=len(z))
                 ):
                     # Skip samples with large forces
@@ -273,7 +273,7 @@ class Ace(MemmappedDataset):
 
                     # Create a sample
                     args = dict(
-                        z=z, pos=pos, y=y.view(1, 1), neg_dy=neg_dy, total_charge=q, partial_charges=pq, dipole_moment=dp
+                        z=z, pos=pos, y=y.view(1, 1), neg_dy=neg_dy, total_charge=total_charge, partial_charges=partial_charges, dipole_moment=dipole_moment
                     )
                     if mol_ids:
                         args["mol_id"] = mol_id
--- a/torchmdnet/datasets/hdf.py
+++ b/torchmdnet/datasets/hdf.py
@@ -57,6 +57,15 @@ class HDF5(Dataset):
                             self.fields.append(
                                 ("partial_charges", "partial_charges", torch.float32)
                             )
+                        # total charge and spin, will be load as 'q' and 's' respectively to keep the same naming convention
+                        if "total_charge" in group:
+                            self.fields.append(
+                                ("total_charge", "total_charge", torch.float32)
+                            )
+                        if "spin" in group:
+                            self.fields.append(
+                                ("spin", "spin", torch.float32)
+                            )
                         assert ("energy" in group) or (
                             "forces" in group
                         ), "Each group must contain at least energies or forces"
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -387,25 +387,14 @@ class TorchMD_Net(nn.Module):
 
         if self.derivative:
             pos.requires_grad_(True)
-        
-        if self.representation_model.additional_methods is None:
-            extra_args_nnp = None 
-        else:
-            extra_args_nnp = {}
-            # force the label to be atom wise
-            for label, t in extra_args.items():
-                if label in self.representation_model.additional_methods.keys():
-                    if t.shape != z.shape:
-                        t = t[batch]
-                    extra_args_nnp[label] = t
-                    
+                   
         # run the potentially wrapped representation model
         x, v, z, pos, batch = self.representation_model(
             z,
             pos,
             batch,
             box=box,
-            extra_args=extra_args_nnp,
+            extra_args=extra_args if self.representation_model.additional_methods else None,
         )
         # apply the output network
         x = self.output_model.pre_reduce(x, v, z, pos, batch)
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -379,7 +379,7 @@ class TorchMD_Net(nn.Module):
         batch = torch.zeros_like(z) if batch is None else batch
 
         # trick to incorporate SPICE pqs
-        # set charge: true in yaml
+        # set charge: true in yaml ((?) currently I do it)
         q = extra_args["pq"]
         
         if self.derivative:
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -348,7 +348,6 @@ class TorchMD_Net(nn.Module):
         batch: Optional[Tensor] = None,
         box: Optional[Tensor] = None,
         extra_args: Optional[Dict[str, Tensor]] = None,
-        extra_fields: Optional[Dict[str, Any]] = None,
     ) -> Tuple[Tensor, Tensor]:
         """
         Compute the output of the model.
@@ -378,8 +377,7 @@ class TorchMD_Net(nn.Module):
             The vectors defining the periodic box.  This must have shape `(3, 3)`,
             where `box_vectors[0] = a`, `box_vectors[1] = b`, and `box_vectors[2] = c`.
             If this is omitted, periodic boundary conditions are not applied.
-            extra_args (Dict[str, Tensor], optional): Extra arguments to pass to the prior model or to the representation model.
-            extra_fields (Dict[str, Tensor], optional): Extra fields to pass to the representation model.
+            extra_args (Dict[str, Tensor], optional): Extra arguments to pass to the prior model and to the representation model.
 
         Returns:
             Tuple[Tensor, Optional[Tensor]]: The output of the model and the derivative of the output with respect to the positions if derivative is True, None otherwise.
@@ -389,27 +387,14 @@ class TorchMD_Net(nn.Module):
 
         if self.derivative:
             pos.requires_grad_(True)
-            
-        # recover the extra_fields_args from the extra_fields
-        if self.representation_model.extra_fields is None:
-            extra_fields_args = None
-        else:
-            assert extra_args is not None, "Extra fields are required but not provided."
-            extra_fields_args = {}
-            for field in extra_fields.keys():
-                t = extra_args[field]
-                if t.shape != z.shape:
-                    # expand molecular label to atom labels
-                    t = t[batch]
-                extra_fields_args[field] = t    
-                    
+        
         # run the potentially wrapped representation model
         x, v, z, pos, batch = self.representation_model(
             z,
             pos,
             batch,
             box=box,
-            extra_fields_args=extra_fields_args,
+            extra_args=extra_args,
         )
         # apply the output network
         x = self.output_model.pre_reduce(x, v, z, pos, batch)
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -47,7 +47,7 @@ def create_model(args, prior_model=None, mean=None, std=None):
         additional_labels = args["additional_labels"]
     else:
         additional_labels = None
-        warnings.warn("Extra fields should be a dictionary. Ignoring extra fields.")
+        warnings.warn("Additional labels should be a dictionary. Ignoring additional labels.")
         
     shared_args = dict(
         hidden_channels=args["embedding_dimension"],
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -388,13 +388,22 @@ class TorchMD_Net(nn.Module):
         if self.derivative:
             pos.requires_grad_(True)
         
+        if self.representation_model.additional_methods is not None:
+            extra_args_nnp = {}
+            # force the label to be atom wise
+            for label, t in extra_args.items():
+                if label in self.representation_model.additional_methods.keys():
+                    if t.shape != z.shape:
+                        t = t[batch]
+                    extra_args_nnp[label] = t
+                    
         # run the potentially wrapped representation model
         x, v, z, pos, batch = self.representation_model(
             z,
             pos,
             batch,
             box=box,
-            extra_args=extra_args,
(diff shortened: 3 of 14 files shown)


tests/test_examples.py
```diff
@@ -27,4 +28,4 @@ def test_example_yamls(fname):
 
     z, pos, batch = create_example_batch()
     model(z, pos, batch)
-    model(z, pos, batch, q=None, s=None)
```
RaulPPelaez, line 30: Should the case `extra_args={"total_charge": None}` be supported?

torchmdnet/datasets/ace.py
```diff
@@ -47,7 +47,7 @@ class Ace(MemmappedDataset):
     - `positions`: Atomic positions. Units: Angstrom.
     - `forces`: Forces on the atoms. Units: eV/Å.
     - `partial_charges`: Atomic partial charges. Units: electron charges.
-    - `dipole_moment` (version 1.0) or `dipole_moments` (version 2.0): Dipole moment (a vector of three components). Units: e*Å.
+    - `dipole_moment` (version 1.0) or `dipole_moment` (version 2.0): Dipole moment (a vector of three components). Units: e*Å.
```
RaulPPelaez, line 50: Are you sure the "s" should not be there?
Not it just says dipole_moment twice. I probably made that distinction in the docstring for a reason.

torchmdnet/datasets/hdf.py
```diff
@@ -57,6 +57,15 @@ def __init__(self, filename, dataset_preload_limit=1024, **kwargs):
                             self.fields.append(
                                 ("partial_charges", "partial_charges", torch.float32)
                             )
+                        # total charge and spin, will be load as 'q' and 's' respectively to keep the same naming convention
```
RaulPPelaez, line 60: This comment does not seem to be enforced later on.
What happens if I have an old hdf around with "q" and "qs" fields?

torchmdnet/models/model.py
```diff
@@ -38,7 +38,17 @@ def create_model(args, prior_model=None, mean=None, std=None):
...
+    # Here we introduce the extra_fields_args, which is Dict[str, Any]
+    # This could be used from each model to initialize nn.embedding layers, nn.Parameter, etc.
+    if "additional_labels" not in args:
+        additional_labels = None
+    elif isinstance(args["additional_labels"], dict):
+        additional_labels = args["additional_labels"]
+    else:
+        additional_labels = None
+        warnings.warn("Additional labels should be a dictionary. Ignoring additional labels.")
+
```
RaulPPelaez, line 51:
```python
additional_labels = args.get("additional_labels")
# Should this check not just be an error?
if additional_labels is not None and not isinstance(additional_labels, dict):
  additional_labels = None
   warnings.warn("Additional labels should be a dictionary. Ignoring additional labels.")
   ```

File: torchmdnet/models/model.py [hunk 5]
```diff
@@ -263,8 +274,8 @@ class TorchMD_Net(nn.Module):
     Parameters
     ----------
     representation_model : nn.Module
-        A model that takes as input the atomic numbers, positions, batch indices, and optionally
-        charges and spins. It must return a tuple of the form (x, v, z, pos, batch), where x
+        A model that takes as input the atomic numbers, positions, batch indices.
```
RaulPPelaez, line 277: The model can also optionally take basically any additional_labels now, right?

torchmdnet/models/model.py
```diff
@@ -380,9 +387,14 @@ def forward(
...
             pos.requires_grad_(True)
+                   
         # run the potentially wrapped representation model
         x, v, z, pos, batch = self.representation_model(
-            z, pos, batch, box=box, q=q, s=s
+            z,
+            pos,
+            batch,
+            box=box,
+            extra_args = extra_args if self.representation_model.additional_methods is not None else None,
```
RaulPPelaez, line 397: You are assuming all representation models have a member called "additional_methods", which I do not think is a good idea. This is an internal implementation detail being leaked.
Why not sending extra_args all the time to the model? It is free to ignore them.

torchmdnet/models/tensornet.py
```diff
@@ -163,6 +169,17 @@ def __init__(
...
+        self.additional_labels = additional_labels
+        # initialize additional methods as None if not provided, also used by module.py 
+        self.additional_methods = None
+        
+        if additional_labels is not None:
+            self.additional_methods = {}
+            for method_name, method_args in additional_labels.items():
+                # the key of the additional_methods is the label of the method (total_charge, partial_charges, etc.)
+                # this will be useful for static shapes processing if needed
+                self.additional_methods[method_args['label']] = {'name': method_name, 'method':self.initialize_additional_method(method_name, method_args)}
```
RaulPPelaez, line 181: I do not really have a better alternative for you, but the names "additional_method" and "initialize_additional_method" should be a little less generic I think.
Maybe "label_handlers" or "additional_label_method"?

torchmdnet/models/tensornet.py
```diff
@@ -213,6 +229,12 @@ def __init__(
 
         self.reset_parameters()
 
+    def initialize_additional_method(self, method, args):
```
RaulPPelaez, line 232: This does not have to be a method of the class, you can have it as a free standing function (like "vector_to_skew_tensor")

torchmdnet/models/tensornet.py
```diff
@@ -236,16 +257,22 @@ def forward(
...
-        else:
-            q = q[batch]
         zp = z
+        if extra_args is not None:
+            # we are assuming that extra args will be used, see model.py forward method how extra_args is passed
+            for label, t in extra_args.items():
+                # molecule wise --> atom wise 
+                if t.shape != z.shape:
+                    extra_args[label] = t[batch]
+
```
RaulPPelaez, line 267: You should not assume everything will be used. You know the additional labels from the initializer.
This loop should be like:
```python
if self.additional_labels is not None:
     assert extra_args is not None
for label in self.additional_labels:
      assert label in extra_args, f"TensorNet expects {label} to be provided as part of extra_args"
      if extra_args[label].shape != z.shape:
         extra_args[label] = extra_args[label][batch]
```

torchmdnet/models/tensornet.py
```diff
@@ -236,16 +257,22 @@ def forward(
...
+                if t.shape != z.shape:
+                    extra_args[label] = t[batch]
+            
         if self.static_shapes:
             mask = (edge_index[0] < 0).unsqueeze(0).expand_as(edge_index)
             zp = torch.cat((z, torch.zeros(1, device=z.device, dtype=z.dtype)), dim=0)
-            q = torch.cat((q, torch.zeros(1, device=q.device, dtype=q.dtype)), dim=0)
+            if self.additional_labels is not None:
+                for label in self.additional_methods.keys():
+                    assert label in extra_args, f"Extra field {label} not found in extra_args"
```
RaulPPelaez, line 273: This assert should go where my previous comment is

torchmdnet/models/tensornet.py
```diff
@@ -480,18 +510,39 @@ def forward(
...
+        X = X + dX + (prefactor) * torch.matrix_power(dX, 2)
         return X
+
+
+class TensornetQ(nn.Module):
+  def __init__(self, init_value, additional_label='total_charge', learnable=False):
+    super().__init__()
+    self.prmtr = nn.Parameter(torch.tensor(init_value), requires_grad=learnable)
+    self.allowed_labels = ['total_charge', 'partial_charges']
+    assert additional_label in self.allowed_labels, f"Label {additional_label} not allowed for this method"
```
RaulPPelaez, line 545: Is there anything in particular preventing this method to be used with any other label apart from these two?
I could see it restricted to "just scalars", but other than that the method seems general.

torchmdnet/models/tensornet.py
```diff
@@ -3,7 +3,7 @@
 # (See accompanying file README.md file or copy at http://opensource.org/licenses/MIT)
 
 import torch
-from typing import Optional, Tuple
+from typing import Optional, Tuple, Dict, Any
```
RaulPPelaez, line 6: Any is not being used.

torchmdnet/models/torchmd_et.py
```diff
@@ -79,7 +79,12 @@ class TorchMD_ET(nn.Module):
             (default: :obj:`False`)
         check_errors (bool, optional): Whether to check for errors in the distance module.
             (default: :obj:`True`)
-
+        additional_labels (Dict[str, Any], optional): Define the additional method to be used by the model, and the parameters to initialize it.
+            example:
+            additional_labels = {method_name1: {label_name: 'extra_arg_label', 'method_prm1': method_prm1,  'method_prm2': method_prm2}, 
+                                 method_name2: {label_name: 'extra_arg_label', 'method_prm1': method_prm1,  'method_prm2': method_prm2},
+                                 ...}
+            (default: :obj:`None`)
```
RaulPPelaez, line 87: I believe there should be no mention to additional_labels whatsoever in ET. Unless you go ahead and implement an actual method (like peters embedding for instance). As it is just now it is just misleading, no matter what I send in this argument it is just going to be silently ignored.
I believe you should at least implement a check of the method names sent to the representation models. For instance, if "TensornetQ" is received by ET that should be an error.
I understand you might want to add this argument for opening the door to future ways of handling the additional labels, but at least you should do something like:
```python
assert additional_labels is None, "ET does not support this feature"
```
The docstring should also reflect this current state of things.

torchmdnet/models/torchmd_gn.py
```diff
@@ -86,6 +86,11 @@ class TorchMD_GN(nn.Module):
             (default: :obj:`None`)
         check_errors (bool, optional): Whether to check for errors in the distance module.
             (default: :obj:`True`)
+        additional_labels (Dict[str, Any], optional): Define the additional method to be used by the model, and the parameters to initialize it.
+            example:
+            additional_labels = {method_name1: {label_name: 'extra_arg_label', 'method_prm1': method_prm1,  'method_prm2': method_prm2}, 
+                                 method_name2: {label_name: 'extra_arg_label', 'method_prm1': method_prm1,  'method_prm2': method_prm2},
+                                 ...}
```
RaulPPelaez, line 93: Same concerns here

torchmdnet/models/torchmd_t.py
```diff
@@ -76,7 +76,11 @@ class TorchMD_T(nn.Module):
             (default: :obj:`None`)
         check_errors (bool, optional): Whether to check for errors in the distance module.
             (default: :obj:`True`)
-
+        additional_labels (Dict[str, Any], optional): Define the additional method to be used by the model, and the parameters to initialize it.
+            example:
+            additional_labels = {method_name1: {label_name: 'extra_arg_label', 'method_prm1': method_prm1,  'method_prm2': method_prm2}, 
+                                 method_name2: {label_name: 'extra_arg_label', 'method_prm1': method_prm1,  'method_prm2': method_prm2},
+                                 ...}
```
RaulPPelaez, line 83: T is marked as deprecated and will be removed at some point, I do not think it is worth the dev effort of adding this feature.

tests/test_examples.py
AntonioMirarchi: I don't think so. If that `additional_label` it's None then it's like I don't have it in my data and moreover I cannot train or do inference with None.

torchmdnet/models/tensornet.py
```diff
@@ -163,6 +169,17 @@ def __init__(
         self.activation = activation
         self.cutoff_lower = cutoff_lower
         self.cutoff_upper = cutoff_upper
+        self.additional_labels = additional_labels
+        # initialize additional methods as None if not provided, also used by module.py
```
RaulPPelaez, line 173: I think this comment is a lie, I do not see additional methods in module.py (which is how it should be)

torchmdnet/module.py
```diff
@@ -8,7 +8,7 @@
 from torch.optim.lr_scheduler import ReduceLROnPlateau
 from torch.nn.functional import local_response_norm, mse_loss, l1_loss
 from torch import Tensor
-from typing import Optional, Dict, Tuple
+from typing import Optional, Dict, Tuple, Any
```
RaulPPelaez, line 11: Any is not used.

torchmdnet/scripts/train.py
```diff
@@ -77,8 +78,6 @@ def get_argparse():
     parser.add_argument('--prior-model', type=str, default=None, help='Which prior model to use. It can be a string, a dict if you want to add arguments for it or a dicts to add more than one prior. e.g. {"Atomref": {"max_z":100}, "Coulomb":{"max_num_neighs"=100, "lower_switch_distance"=4, "upper_switch_distance"=8}', action="extend", nargs="*")
 
     # architectural args
-    parser.add_argument('--charge', type=bool, default=False, help='Model needs a total charge. Set this to True if your dataset contains charges and you want them passed down to the model.')
-    parser.add_argument('--spin', type=bool, default=False, help='Model needs a spin state. Set this to True if your dataset contains spin states and you want them passed down to the model.')
```
RaulPPelaez, line 81: This could break previous yamls. Maybe it is better to leave them for now, marked as deprecated and document them as "This option is obsolete and ignored".

torchmdnet/datasets/ace.py
AntonioMirarchi: Yes you're right. The layout version 2.0 of ace it's using `dipole_moments`. I considered that in the code but I changed the name in the documentation.

RaulPPelaez:
It would be wise to add a new test file in which you poke the new functionality around, for instance making sure that the input is being received correctly or that the output of the model changes depending on whether additional_labels is passed or not.

torchmdnet/datasets/hdf.py
AntonioMirarchi: You should open it and rename the datasets with the allowed labels, i.e. `total_charge` and `spin`. Otherwise this dataloader will be if/elif/else

RaulPPelaez: But this is contradicting the comment, which says that those properties will be loaded as "q" and "s"?

AntonioMirarchi: yes, it's a residue. I'm removing it

torchmdnet/models/model.py
AntonioMirarchi: Each model has `additional_methods` defined in its initialization. I was thinking about it at some point, but then I need in any case a way in the forward method of each architecture to do some assertion and pre-processing on the data. So at the end, I decided to use as common line `if extra_args is not None:` in the forward of the architecture (see [this ](https://github.com/torchmd/torchmd-net/blob/2b18f905f51c118ac1bcf30119bc6d02e4fd0669/torchmdnet/models/tensornet.py#L261)for example)

torchmdnet/models/model.py
```diff
@@ -3,7 +3,7 @@
 # (See accompanying file README.md file or copy at http://opensource.org/licenses/MIT)
 
 import re
-from typing import Optional, List, Tuple, Dict
+from typing import Optional, List, Tuple, Dict, Any
```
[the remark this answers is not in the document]
AntonioMirarchi: Nope, I will remove it.

torchmdnet/models/tensornet.py
AntonioMirarchi: True, fixed!

torchmdnet/models/tensornet.py
AntonioMirarchi: Mmmm not sure, repeating additional_labels too many times is redundant. `additional_methods` handles only additional_labels, the relationship seems quite straightforward and I would say it is quite generic

torchmdnet/models/tensornet.py
AntonioMirarchi: Yes, this was correctly implemented before when I was defining which extra_args should go to the forward pass in model.py. I will fix it, thanks!

torchmdnet/models/tensornet.py
AntonioMirarchi: Yes, I will also move the `static_shapes` assertion and reshaping above, so the for loop on the additional_labels will only take place once

torchmdnet/models/model.py
RaulPPelaez: My point is architectures should not require to have a member called `additional_methods`. And external classes should not depend on it or access it in general.
Both parties (TorchMD_Net and the representation models) have enough information on their own. 
TorchMD_Net should just send extra_args all the time, you are assuming extra_args will be used exclusively for your new additional labels feature, but this might not be the case at all. I could implement it so that I can send in extra_args something like "check_errors", for instance.
OTOH the representation models have several members that they can use to know what to use from extra_args, if anything.

torchmdnet/models/tensornet.py
RaulPPelaez: The fact that "additional_methods" handles only additional labels is the problem. The name carries too little meaning of what the variable does. It is akin to calling a member variable "additional_variable".

torchmdnet/scripts/train.py
AntonioMirarchi: Are you sure about it? In theory the test_examples is using yaml in which `charge` and `spin` are defined but it's not complaining about it because the code it's just skipping and ignoring them.

torchmdnet/module.py
AntonioMirarchi: removed

torchmdnet/models/tensornet.py
RaulPPelaez: You could also get rid of the variable altogether and store everything in additional_labels, or move it to a separate class.
The way you are using it now reminds me of callbacks, you could all it "label_callbacks"

torchmdnet/models/tensornet.py
AntonioMirarchi: Yes sorry it's model.py. It's used in the forward step in model.py to define if extra_args should be None. Anyway  I will remove it, since now all the extra_args will be passed in any case and then the assertion will be inside the forward of each model.

torchmdnet/models/torchmd_t.py
AntonioMirarchi: ok but if someone wants do it, it's there

torchmdnet/models/torchmd_gn.py
AntonioMirarchi: same answer as for the transformer

torchmdnet/models/torchmd_t.py
RaulPPelaez: That is fine, but at least reflect that it is not allowed to send it now and make the code fail if it is not None. IMO this is just adding additional unnecessary clutter to the code. 
It would be better to have a section in the documentation in the part about adding a new representation model saying "How to handle additional labels"

torchmdnet/models/model.py
AntonioMirarchi: ending extra_args all the time to the model according also to the other comments in this PR

torchmdnet/models/tensornet.py
AntonioMirarchi: My original idea was to reintroduce only the additional_labels available now with this PR, but in theory the input could be anything

torchmdnet/models/tensornet.py
```diff
@@ -480,18 +510,39 @@ def forward(
             edge_index, edge_attr[..., 2, None, None], S, X.shape[0]
         )
         msg = Im + Am + Sm
+        
+        prefactor = 1 if self.addtional_methods is not None else torch.ones_like(msg, device=msg.device, dtype=msg.dtype)
```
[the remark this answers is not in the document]
AntonioMirarchi: currently is 1 when additional_labels is not None. But if I set is as 1 and additional_labels is None, this will return the normal behavior of tensornet

torchmdnet/models/tensornet.py
```diff
@@ -480,18 +510,39 @@ def forward(
...
+        if self.addtional_methods is not None and extra_args is not None:
+            for label, method_dict in self.addtional_methods.items():
+                # appending to this list all the methods will be working in this way
```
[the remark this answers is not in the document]
AntonioMirarchi: This comment it's linked to the #TODO defined later. So, considering a case in which you want to train with total_charge and partal_charges, then how do you want to add this information on the prefactor. Simple sum or maybe appending to a list (or array) and multiply them before then adding to the prefactor? @guillemsimeon

torchmdnet/models/tensornet.py
```diff
@@ -480,18 +510,39 @@ def forward(
             edge_index, edge_attr[..., 2, None, None], S, X.shape[0]
         )
         msg = Im + Am + Sm
+        
+        prefactor = 1 if self.addtional_methods is not None else torch.ones_like(msg, device=msg.device, dtype=msg.dtype)
+        if self.addtional_methods is not None and extra_args is not None:
+            for label, method_dict in self.addtional_methods.items():
```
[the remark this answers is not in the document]
AntonioMirarchi: not really, it's a simple loop on a dict and you can access tensornetq only with the related additonal_label stored in additional_dict. Moreover the loop let you to consider multiple additonal_labels working with tensornetQ

torchmdnet/models/tensornet.py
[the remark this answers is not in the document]
RaulPPelaez: Yes, it implements an equivalent behavior, but you are allocating a really big tensor to just fill it with ones and multiply it elementwise, which is expensive. My point here is you should be able to just set it to 1 always. Or make it a size 1 tensor that you expand_as(whatever)

torchmdnet/models/torchmd_et.py
AntonioMirarchi: I will add that assertion to all the models which are not using additional_labels for now. The check on the method name it's done by the initialize_additional_method function (like the tensornet one). I can add it for each model if you want but at the moment since any additional_method is available (except for tensornet) I would say the assertion it's fine

torchmdnet/models/tensornet.py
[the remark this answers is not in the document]
RaulPPelaez: I understand the expected length of this additional_methods list is short, but in general one tries to do as much preprocessing as possible at construction.
additional_methods is not supposed to change after initialization, so you could create:
```python
self.tensorq_labels = [method_and_whatever for method in self.additional_methods if method["name"]=="tensorq"]
```
So then you can just do:
```python
for label, callback in self.tensorq_labels:
    labels_broadcasted = extra_args[label][..., None, None, None]
    prefactor += labels_broadcasted
```

torchmdnet/scripts/train.py
AntonioMirarchi: No, you cannot. The test_examples it's just ignoring the check called in `LoadFromFile(argparse.Action)`. I will put them back and marked as deprecated

torchmdnet/models/tensornet.py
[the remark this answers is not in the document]
AntonioMirarchi: OK, I will do something like that but `labels_broadcasted` will be directly the output of tensornetQ

torchmdnet/models/tensornet.py
AntonioMirarchi: okok let's move to label_callbacks

torchmdnet/models/tensornet.py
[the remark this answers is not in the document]
AntonioMirarchi: using torch.prod and torch.stack for the moment: [here](https://github.com/torchmd/torchmd-net/blob/6b72a00b51e62dd699729af1bc11cec75f6963fd/torchmdnet/models/tensornet.py#L537)

PR #308, opened 2024-03-18
Added support for ensemble models

stefdoerr:
Works fine but I don't like so much that the same function can return a model with 4 outputs (energy/force mean/std) or with 2 outputs (just energy, force) depending on if you pass a single file or multiple.
Maybe it could become an argument to load_model to disable the returning of the std?

### DIFF
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -434,9 +434,17 @@ class TorchMD_Net(nn.Module):
 
 
 class Ensemble(torch.nn.ModuleList):
-    """Average predictions over an ensemble of TorchMD-Net models"""
+    """Average predictions over an ensemble of TorchMD-Net models.
 
-    def __init__(self, modules):
+       This module behaves like a single TorchMD-Net model, but its forward method returns the average and standard deviation of the predictions over all models it was initialized with.
+
+    Args:
+        modules (List[nn.Module]): List of :py:mod:`TorchMD_Net` models to average predictions over.
+    """
+
+    def __init__(self, modules: List[nn.Module]):
+        for module in modules:
+            assert isinstance(module, TorchMD_Net)
         super().__init__(modules)
 
     def forward(
--- a/torchmdnet/models/model.py
+++ b/torchmdnet/models/model.py
@@ -151,6 +151,10 @@ def load_model(filepath, args=None, device="cpu", **kwargs):
     Returns:
         nn.Module: An instance of the TorchMD_Net model.
     """
+    if isinstance(filepath, (list, tuple)):
+        return Ensemble(
+            [load_model(f, args=args, device=device, **kwargs) for f in filepath]
+        )
 
     ckpt = torch.load(filepath, map_location="cpu")
     if args is None:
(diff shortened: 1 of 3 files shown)


RaulPPelaez:
Depending on your final use of this you could return a Dict[str, Tensor] or a Tuple[Tuple[Tensor, Optional[Tensor]], Tuple[Tensor, Optional[Tensor]]] or a List[Dict[str, Optional[Tensor]]] storing mean and std for each output

torchmdnet/models/model.py
```diff
@@ -426,3 +430,37 @@ def forward(
...
+
+    def forward(
+        self,
+        z: Tensor,
+        pos: Tensor,
+        batch: Optional[Tensor] = None,
+        box: Optional[Tensor] = None,
+        q: Optional[Tensor] = None,
+        s: Optional[Tensor] = None,
+        extra_args: Optional[Dict[str, Tensor]] = None,
```
RaulPPelaez, line 449: It would be interesting to make this class take *args,**kwargs for future proofing. I do not think it is compatible with TorchScript, tough.

torchmdnet/models/model.py
```diff
@@ -426,3 +430,37 @@ def forward(
...
+            y.append(res[0])
+            neg_dy.append(res[1])
+
+        y = torch.stack(y)
+        neg_dy = torch.stack(neg_dy)
+        y_mean = torch.mean(y, axis=0)
+        neg_dy_mean = torch.mean(neg_dy, axis=0)
+        y_std = torch.std(y, axis=0)
+        neg_dy_std = torch.std(neg_dy, axis=0)
+        return y_mean, neg_dy_mean, y_std, neg_dy_std
```
RaulPPelaez, line 466: You could change it to `{"y": {"mean": y_mean, "std": y_std}, "neg_dy":{"mean": neg_dy_mean, "std": neg_dy_std}}` 
It would be interesting to have this class be somehow interface-compatible with TorchMD_Net, but you are returning 4 tensors anyway, so it cannot be.

RaulPPelaez:
I believe your implementation is clean enough for its purposes. Stacking the model outputs is wasteful, but the code is much cleaner and adaptable than a Welford-like an online computation.

Lets settle on the output and we can write a documentation entry for it and some tests. Maybe below this entry? https://torchmd-net.readthedocs.io/en/latest/models.html#loading-a-model-for-inference

RaulPPelaez:
I added a sanity check:
```python
def test_ensemble():
    ckpts = [join(dirname(dirname(__file__)), "tests", "example.ckpt")] * 3
    model = load_model(ckpts[0])
    ensemble_model = load_model(ckpts)
    z, pos, batch = create_example_batch(n_atoms=5)

    pred, deriv = model(z, pos, batch)
    pred_ensemble, deriv_ensemble, y_std, neg_dy_std = ensemble_model(z, pos, batch)

    torch.testing.assert_close(pred, pred_ensemble, atol=1e-5, rtol=1e-5)
    torch.testing.assert_close(deriv, deriv_ensemble, atol=1e-5, rtol=1e-5)
    assert y_std.shape == pred.shape
    assert neg_dy_std.shape == deriv.shape
    assert (y_std == 0).all()
    assert (neg_dy_std == 0).all()
```

torchmdnet/models/model.py
stefdoerr: Yes I definitely considered that. But if you say it's not compatible we can leave it for now I guess

stefdoerr:
Nice sanity test, thanks! Okay so we can merge it for now and we see in the future if it needs adaptation for different model types

torchmdnet/models/model.py
stefdoerr: I think I prefer it as is in the sense that it keeps compatibility in the first 2 returned arguments. So all models return as first the y and second the neg_dy and this one does too. And the other args are extra information. Converting the result to a dict will change more the style I think.

stefdoerr: I think if I can add an argument to load_model called `return_std=True` then it would be clearer when the Ensemble returns two or 4 args. I'll do that

stefdoerr: I implemented this now. Give it another look

torchmdnet/models/model.py
```diff
@@ -426,3 +433,51 @@ def forward(
...
+        neg_dy = torch.stack(neg_dy)
+        y_mean = torch.mean(y, axis=0)
+        neg_dy_mean = torch.mean(neg_dy, axis=0)
+        y_std = torch.std(y, axis=0)
+        neg_dy_std = torch.std(neg_dy, axis=0)
+
+        if self.return_std:
+            return y_mean, neg_dy_mean, y_std, neg_dy_std
+        else:
+            return y_mean, neg_dy_mean
```
RaulPPelaez, line 483: Not sure if you care about that but this is probably not TorchScript compatible just as is.

stefdoerr: Ah true. Where do we use TorchScript generally? So that I see if it affects us

RaulPPelaez: Whenever you need to serialize the model. Aka torch.load/torch.save.

stefdoerr: We will probably not need that, no? Considering that we train individual models and this is a way to load a number of them together into an ensemble for evaluation only.

torchmdnet/models/model.py
RaulPPelaez: Given that we are not aiming for TorchScript anymore, I would make this function variadic so that it is immune to changes in the TorchMD_Net interface.
This is particularly useful for when/if #306 is merged, which changes how we deal with charges, spins and such.

RaulPPelaez (approved):

stefdoerr:
Looks lovely. Many thanks Raul!

PR #311, opened 2024-03-27
Support for ensemble model zip files

stefdoerr:
Allow also passing a single zip file containing many ckpt instead of a list of ckpt files


RaulPPelaez:
Load_model was getting a bit cluttered so I moved the Ensemble loading functionality to a different function. Also added some assertions and updated the docs.

I think it is fine to have the imports globally, everything is in python proper.

Feel free to add/disregard and merge yourself when you are ready.
Thanks!

RaulPPelaez (approved):

PR #314, opened 2024-04-04
Allow to configure the depth of the MLP in output modules 

RaulPPelaez:
Adds an MLP module to utils and uses it in the output models.

Allows to configure the depth of it via a train.py option, `output-mlp-num-layers`

### DIFF
--- a/torchmdnet/models/utils.py
+++ b/torchmdnet/models/utils.py
@@ -434,6 +434,49 @@ class CosineCutoff(nn.Module):
             return cutoffs
 
 
+class MLP(nn.Module):
+    """A simple multi-layer perceptron with a given number of layers and hidden channels.
+
+    Args:
+        in_channels (int): Number of input features.
+        out_channels (int): Number of output features.
+        hidden_channels (int): Number of hidden features.
+        activation (str): Activation function to use.
+        num_layers (int, optional): Number of layers. Defaults to 0.
+        dtype (torch.dtype, optional): Data type to use. Defaults to torch.float32.
+    """
+
+    def __init__(
+        self,
+        in_channels,
+        out_channels,
+        hidden_channels,
+        activation,
+        num_layers=0,
+        dtype=torch.float32,
+    ):
+        super(MLP, self).__init__()
+        act_class = act_class_mapping[activation]
+        self.act = act_class()
+        self.layers = nn.Sequential()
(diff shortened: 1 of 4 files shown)


torchmdnet/models/utils.py
```diff
@@ -434,6 +434,49 @@ def forward(self, distances: Tensor) -> Tensor:
...
+        dtype (torch.dtype, optional): Data type to use. Defaults to torch.float32.
+    """
+
+    def __init__(
+        self,
+        in_channels,
+        out_channels,
+        hidden_channels,
+        activation,
+        num_layers=0,
```
stefdoerr, line 455: Default before was 2, no?

RaulPPelaez: Actually this might be a misconception on my part regarding naming.
num_layers=0 means:
act(linear): in_channels -> hidden_channels
linear: hidden_channels -> out_channels

num_layers=1 would be:
act(linear): in_channels -> hidden_channels
act(linear): hidden_channels->hidden_channels
linear: hidden_channels -> out_channels

It makes sense to me because otherwise num_layers<2 would not be possible. But maybe my naming is not the standard?

guillemsimeon: I like this convention:

num_layers=1 would be:
act(linear): in_channels -> hidden_channels
act(linear): hidden_channels->hidden_channels
linear: hidden_channels -> out_channels

But I would call them hidden layers, instead of layers.

guillemsimeon: The first one is the input layer, the last one the output layer.

RaulPPelaez: Changed the name to num_hidden_layers and clarified in the docstring.

torchmdnet/scripts/train.py
```diff
@@ -74,6 +74,7 @@ def get_argparse():
     # model architecture
     parser.add_argument('--model', type=str, default='graph-network', choices=models.__all_models__, help='Which model to train')
     parser.add_argument('--output-model', type=str, default='Scalar', choices=output_modules.__all__, help='The type of output model')
+    parser.add_argument('--output-mlp-num-layers', type=int, default=0, help='If the output model uses an MLP this will be the number of inner layers.')
```
guillemsimeon, line 77: I would say: number of hidden layers, excluding input and output layers.

guillemsimeon (approved):
thanks Raul, it was high time to have a MLP direct implementation :)

PR #316, opened 2024-04-11
Filter non used sample values in QM9

RaulPPelaez:
Samples provided by the QM9 dataset included a bunch of unused tensors, like "edge_index".
This was slowing down training a bit by adding overhead when building batches.
This PR removes them via a pre_transform, leaving only "pos", "z" and "y".


guillemsimeon (approved):
lgtm

PR #317, opened 2024-04-17
Train only an output model, freezing the representation model.

RaulPPelaez:


RaulPPelaez:
cc @stefdoerr

RaulPPelaez:
@stefdoerr, note that this allows only to train models with identical Output heads (in terms of shape and type). If we want to support something more generic, like having different heads in different trainings (or the same with different shapes), we could instead use the following approach:
- Create the model as in a regular training, which allows to fabricate any combination of repr + output.
- Add an option to overwrite the initial weights of the representation model and then freeze them.
I believe this strategy is more generic, it should be done in this way in the final implementation.

PR #319, opened 2024-05-08
Enable allow_pickle to be able to specify splits.

brian8128:
I've been trying to train a torchmd-net model following the examples in the examples folder but I'm unable to specify my own splits. This is an easy fix that unblocks me and perhaps other users. If you know of a way to save the splits in a way that doesn't require us to enable allow_pickle I'd love to hear about it and we can close the PR. 

Full disclosure about the level of testing - this change allows me to get past the part of the code where the splits are loaded but I'm currently hitting other errors later, so I can't yet verify that the full script runs.

### DIFF
--- a/torchmdnet/utils.py
+++ b/torchmdnet/utils.py
@@ -247,7 +247,7 @@ def make_splits(
     order=None,
 ):
     if splits is not None:
-        splits = np.load(splits)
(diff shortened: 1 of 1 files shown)


stefdoerr (approved):
Looks ok to me. Thanks for the contribution!

PR #321, opened 2024-05-13
Exponential moving averages of model parameters

AntonioMirarchi:
Enhance model training by enabling Exponential Moving Average (EMA) on the model parameters. To activate this feature, simply specify the `ema_prmtrs_decay` arg in the configuration file. If not set, its default value is None, indicating that EMA is not going to be applied.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -17,3 +17,5 @@ dependencies:
   - pytest
   - psutil
   - gxx<12
+  - pip:
+    - torch-ema
--- a/torchmdnet/module.py
+++ b/torchmdnet/module.py
@@ -14,6 +14,7 @@ from lightning import LightningModule
 from torchmdnet.models.model import create_model, load_model
 from torchmdnet.models.utils import dtype_mapping
 import torch_geometric.transforms as T
+from torch_ema import ExponentialMovingAverage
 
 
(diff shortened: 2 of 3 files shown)


environment.yml
```diff
@@ -17,3 +17,5 @@ dependencies:
   - pytest
   - psutil
   - gxx<12
+  - pip:
```
RaulPPelaez, line 20: AFAIK this is a dangerous idea. Note that this project you link to is a short and single MIT-licensed class: https://github.com/fadel/pytorch_ema/blob/master/torch_ema/ema.py

I would consider just including it in the repo with the appropriate copyright notice.

torchmdnet/module.py
```diff
@@ -73,6 +74,11 @@ def __init__(self, hparams, prior_model=None, mean=None, std=None):
         else:
             self.model = create_model(self.hparams, prior_model, mean, std)
 
+        self.ema_prmtrs = None
+        if self.hparams.ema_prmtrs_decay is not None:
```
RaulPPelaez, line 78: Tests fail because of this. You have to first make sure the element exists, see how its done with the charge above:
```python
        if "charge" not in hparams:
```
Alternatively you can deal with hparams as a dict before the call to save_hyperparameters.

PR #322, opened 2024-05-16
Disable persistent_workers with num_workers=0 and disable pin_memory in DataModule 

### LINKED ISSUE #323: cannot set num_workers=0
I sometimes get errors related to peristent workers + pinned memory, e.g.:
```
ataloader.py 1290 _get_data
raise RuntimeError('Pin memory thread exited unexpectedly')

RuntimeError:
Pin memory thread exited unexpectedly
[rank: 2] Child process with PID 1662142 terminated with code 1. Forcefully terminating all other processes to avoid zombies 🧟
```
One solution is to set num_workers=0
If I try and do this I get a different error:
```
dataloader.py 254 __init__
raise ValueError('persistent_workers option needs num_workers > 0')

ValueError:
persistent_workers option needs num_workers > 0

```
The solution is in:
#322

### LINKED ISSUE #324: persistent workers=True and Pinned memory=True is unstable
When running with default settings I get a warning:
```
/scratch/users/sfarr/miniconda3/envs/tmd/lib/python3.12/site-packages/lightning/pytorch/trainer/connectors/data_connector.py:448: The combination of `DataLoader(`pin_memory=True`, `persistent_workers=True`) and `Trainer(reload_dataloaders_every_n_epochs > 0)` can lead to instability due to limitations in PyTorch (https://github.com/pytorch/pytorch/issues/91252). We recommend setting `pin_memory=False` in this case.
```

It will run the training for a long time but eventually crash with something like (using 4 GPUs):
```
File "/home/steve/miniconda3/envs/tmd/lib/python3.11/site-packages/torch/utils/data/dataloader.py", line 1289, in _get_data
    raise RuntimeError('Pin memory thread exited unexpectedly')
RuntimeError: Pin memory thread exited unexpectedly
[rank: 1] Child process with PID 224779 terminated with code 1. Forcefully terminating all other processes to avoid zombies :zombie:
```

I can make it stable by setting number workers to zero and #322 
But I can see that the performance is slightly lower on 4 gpus (4.8 it/s vs 5.0 it/s) than if I use num_workers=4 which eventually crashes.

These options should be able to be set in the config yaml. I do not know the performance effects of either.

sef43:
currently I cannot set num_workers to zero:
```
ValueError:
persistent_workers option needs num_workers > 0
```
This was the fix suggested by @AntonioMirarchi that works for me

Closes #323
Closes #324

### DIFF
--- a/torchmdnet/data.py
+++ b/torchmdnet/data.py
@@ -134,7 +134,7 @@ class DataModule(LightningDataModule):
             batch_size=batch_size,
             num_workers=self.hparams["num_workers"],
             persistent_workers=True if self.hparams["num_workers"] > 0 else False,
-            pin_memory=True,
+            pin_memory=False,
             shuffle=shuffle,
         )
 
--- a/torchmdnet/data.py
+++ b/torchmdnet/data.py
@@ -133,7 +133,7 @@ class DataModule(LightningDataModule):
             dataset=dataset,
             batch_size=batch_size,
             num_workers=self.hparams["num_workers"],
-            persistent_workers=True,
+            persistent_workers=True if self.hparams["num_workers"] > 0 else False,
             pin_memory=True,
             shuffle=shuffle,
         )


RaulPPelaez:
Thanks @sef43, I remember setting that option already with @AntonioMirarchi, maybe it got lost in some pending PR.

PR #326, opened 2024-05-17
Update installation.rst

mixarcid:
Seems like this package is only available via conda forge


docs/source/installation.rst
```diff
@@ -5,7 +5,7 @@ TorchMD-Net is available in `conda-forge <https://conda-forge.org/>`_ and can be
 
 .. code-block:: shell
 
-   mamba install torchmd-net
+   mamba install torchmd-net -c conda-forge
```
RaulPPelaez, line 8: Mamba enables conda-forge by default: https://mamba.readthedocs.io/en/latest/installation/mamba-installation.html

PR #329, opened 2024-06-10
Update environment file

RaulPPelaez:
The env file is a little behind, this PR tries to relax the pins in environment.yml.

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -6,9 +6,9 @@ dependencies:
   - matplotlib-base
   - nnpops
   - pip
-  - pytorch<2.2
-  - pytorch_geometric<2.5
-  - lightning<2.2
+  - pytorch
+  - pytorch_geometric
+  - lightning
   - pydantic
   - torchmetrics
   - tqdm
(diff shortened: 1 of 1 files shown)


RaulPPelaez:
Using forward instead of __call__ here:
https://github.com/torchmd/torchmd-net/blob/e224dc500023f52c46429d4a447f5c59d2145b29/torchmdnet/module.py#L19-L32

makes tmdnet incompatible with pytorch_geometric versions older than 2.5. Its a simple change to get both versions to work, so I changed it.

This is BaseTrasnform in 2.5:
https://pytorch-geometric.readthedocs.io/en/latest/_modules/torch_geometric/transforms/base_transform.html#BaseTransform
This is in 2.3:
https://pytorch-geometric.readthedocs.io/en/2.3.0/_modules/torch_geometric/transforms/base_transform.html

RaulPPelaez:
cc @stefdoerr could you please review?

stefdoerr (approved):

PR #331, opened 2024-06-21
Force gradient regularization

RaulPPelaez:
This PR introduces a new optional term to the loss that depends on the gradient of the force.
The hope here is that by penalizing large force gradients the resulting forces will be smoother.

### DIFF
--- a/torchmdnet/module.py
+++ b/torchmdnet/module.py
@@ -167,6 +167,32 @@ class LNNP(LightningModule):
             loss_y = self._update_loss_with_ema(stage, "y", loss_name, loss_y)
         return {"y": loss_y, "neg_dy": loss_neg_y}
 
+    def _compute_second_derivative_regularization(self, y, neg_dy, batch):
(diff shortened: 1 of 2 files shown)


torchmdnet/module.py
```diff
@@ -167,6 +167,32 @@ def _compute_losses(self, y, neg_y, batch, loss_fn, stage):
...
+    def _compute_second_derivative_regularization(self, y, neg_dy, batch):
+        # Compute force gradient and add it to the loss like: max(0, grad(neg_dy.sum())-eps)^2
+        # Args:
+        #   y: predicted value
+        #   neg_dy: predicted negative derivative
+        #   batch: batch of data
+        # Returns:
+        #   regularization: regularization term
+        assert "pos" in batch
+        force_sum = neg_dy.sum()
```
guillemsimeon: This will allways be zero if there are no external forces acting on the molecule, which is always the case in our setting. Internal forces always add up to zero.

PR #333, opened 2024-06-28
Add SPICE 2.0.1, Fix bug in MemmappedDataset

RaulPPelaez:
Adds compatibility for the latest SPICE version https://arxiv.org/pdf/2406.13112

In doing so I uncovered a couple of issues related to MemmapedDataset.
### I fixed MemmapedDataset overwriting the names of processed datasets.
 In particular, SPICE sets self.name as:
```python
        arg_hash = f"{version}{subsets}{max_gradient}{subsample_molecules}"
        arg_hash = hashlib.md5(arg_hash.encode()).hexdigest()
        self.name = f"{self.__class__.__name__}-{arg_hash}"
``` 
Which gets overwritten in the MemmappedDataset constructor:
```python
self.name = self.__class__.__name__
```
Causing any subsample/gradient configurations to be stored as just "SPICE.*.mmap". Probably ignoring subsample/gradient if the dataset has been already processed.

### This in turn uncovered a nasty bug in the test for the ACE Dataset. 
Since all instances of the Dataset are named with just `self.__class__.__name__`, creating two instances of ACE will simply ignore the data in the second and use what was processed in the first to a group of files just called "Ace.*.mmap".
This test should not pass but it was passing until now:
https://github.com/torchmd/torchmd-net/blob/e908988440f2e9d77029a15c41fcee8139ee8b7f/tests/test_datasets.py#L246-L299
In Ace v2 the field is called `dipole_moments`. But MemmappedDataset was not even trying to process the second one.

### I also made some changes
- Now SPICE checks the hash of the downloaded files.
- Now SPICE stores files under `{root}/raw/spice/` or `{root}/processed/spice`. Before it was storing things under `{root}/raw/{version}`, which could collide with other datasets. 

**Note that this PR will invalidate many preprocessed datasets, prompting for redownloading and reprocessing**. I recommend you delete the dataset storage folder and let it redownload things.

### DIFF
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -291,8 +291,8 @@ def test_ace(tmpdir):
         mol["forces"].attrs["units"] = "eV/Å"
         mol["partial_charges"] = np.random.random((2, 3))
         mol["partial_charges"].attrs["units"] = "e"
-        mol["dipole_moment"] = np.random.random((2, 3))
-        mol["dipole_moment"].attrs["units"] = "e*Å"
+        mol["dipole_moments"] = np.random.random((2, 3))
+        mol["dipole_moments"].attrs["units"] = "e*Å"
     dataset_v2 = Ace(root=tmpdir, paths=tmpfilename_v2)
     assert len(dataset_v2) == 6
     f2.flush()
--- a/tests/test_datasets.py
+++ b/tests/test_datasets.py
@@ -279,7 +279,7 @@ def test_ace(tmpdir):
     f2.attrs["layout_version"] = "2.0"
     f2.attrs["name"] = "sample_molecule_data_v2"
     master_mol_group = f2.create_group("master_molecule_group")
-    for m in range(3):  # Three molecules
+    for m in range(4):
         mol = master_mol_group.create_group(f"mol_{m+1}")
         mol["atomic_numbers"] = [1, 6, 8]  # H, C, O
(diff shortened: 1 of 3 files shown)


RaulPPelaez:
There is one bogus molecule in the zenodo hosted 2.0.1 hdf5 file.
```shell
>>> ds = SPICE("~/data/", version="2.0.1")
Downloading https://zenodo.org/records/10975225/files/SPICE-2.0.1.hdf5
Processing...
Gathering statistics...
Molecules: 40119it [01:46, 661.97it/s]WARNING:root:Bogus molecule with id 54X VAL
WARNING:root:Found torch.Size([0]) positions, torch.Size([0]) energies and torch.Size([0]) gradients
```
I downloaded the file a couple times to discard a corruption on my end. Also confirmed the md5 hash coincides with the one in zenodo.
cc @peastman

RaulPPelaez:
@stefdoerr could you please review? this is an important one.

stefdoerr (approved):
ah nice. I was aware of the overwriting and I was just working around it by changing the root dir for each Ace version. But this is of course the much better way of doing it. Thanks!

peastman:
> There is one bogus molecule in the zenodo hosted 2.0.1 hdf5 file.

Thanks!  It's only "bogus" in that there are no conformations for it.  It's a dimer that only had one conformation to begin with, and it looks like that one was excluded because the forces were too large.

PR #360, opened 2025-02-14
Update setup.py to check if WITH_CUDA is set

szaman19:
`os.environ.get("WITH_CUDA", "0") == "0"` return `True` even if the environment variable is not set. Thus currently, the default behavior is to automatically fall back to CPU-only build if `WITH_CUDA` is not set. 

This introduces a small change to the first check if the `WITH_CUDA` is set. If not set, it falls back to the build defined by `torch.cuda._is_compiled()`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,10 +9,13 @@ import os
 
 
 # If WITH_CUDA is defined
-if os.environ.get("WITH_CUDA", "0") == "1":
-    use_cuda = True
-elif os.environ.get("WITH_CUDA", "0") == "0":
-    use_cuda = False
+if os.getenv("WITH_CUDA") is None:
+    if os.environ.get("WITH_CUDA", "0") == "1":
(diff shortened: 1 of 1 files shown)


stefdoerr:
Ah right, of course. I was so focused on making conda-forge work correctly. I will fix it

stefdoerr (approved):

stefdoerr:
I should really not review code before my first morning coffee. There were two logic errors in this PR (the first if was supposed to be `not None`) and there should not be defaults in the nested getenvs. Fixed here:
https://github.com/torchmd/torchmd-net/commit/a302aab9c8f82781e8088b1c2343cb01e185c585

szaman19:
Haha looks like I shouldn't submit a PR before I've had my first coffee either! Thanks for the fix and quick turn around. Cheers!

Issue #372, opened 2025-12-10
Hardcoded PyTorch version


peastman:
The `pyproject.toml` file hardcodes a specific version of PyTorch:

https://github.com/torchmd/torchmd-net/blob/75b16e625039a5e5e79c5e9d679c6c92ee274445/pyproject.toml#L30

Is there a reason that's required?  It prevents installing TorchMD-Net into an environment with any other version of PyTorch, which creates a lot of compatibility roadblocks.

peastman:
That's excellent.  I've seen claims that Triton can be almost as fast as raw CUDA kernels.  I don't know how generally true that is, but if it can work, it seems like a good way to reduce version dependencies.

stefdoerr:
I'm pasting here the results of the benchmarks on an RTX 4090.
Old CUDA implementation

```
Benchmarking neighbor list generation for 32767 particles with 64 neighbors on average
Summary
-------
Batch size Distance(ms)         Brute(ms)            Cell(ms)             
--------------------------------------------------------------------------
1          5.37  x1.00          3.04  x1.77          1.03  x5.19          
2          3.17  x1.00          2.37  x1.34          1.01  x3.14          
4          2.67  x1.00          2.24  x1.19          1.21  x2.20          
8          1.59  x1.00          2.16  x0.74          1.32  x1.21          
16         1.42  x1.00          2.14  x0.66          1.34  x1.05          
32         1.25  x1.00          2.12  x0.59          1.82  x0.69          
64         1.12  x1.00          2.12  x0.53          2.85  x0.39          
128        1.05  x1.00          2.11  x0.50          3.04  x0.35          
256        1.01  x1.00          2.11  x0.48          4.11  x0.25          
512        0.95  x1.00          2.11  x0.45          3.12  x0.30          
Benchmarking neighbor list generation for 1 batches with 64 neighbors on average
Summary
-------
N Particles Distance(ms)         Brute(ms)            Cell(ms)             
---------------------------------------------------------------------------
256         0.55  x1.00          0.01  x62.01         0.14  x3.85          
512         0.58  x1.00          0.01  x53.17         0.19  x3.11          
1024        0.65  x1.00          0.02  x35.45         0.19  x3.35          
2048        0.78  x1.00          0.04  x20.07         0.21  x3.64          
4096        1.05  x1.00          0.09  x11.99         0.26  x3.95          
8192        1.60  x1.00          0.22  x7.44          0.31  x5.14          
[...]

PR #373, opened 2025-12-16
Triton neighbor list implementation

stefdoerr:
Currently installation, maintenance and development of torchmd-net is hampered by the neighbor list CUDA and C++ extensions (see also: https://github.com/torchmd/torchmd-net/issues/372#issuecomment-3642795116). It's very complicated to create and make available python wheels which support multiple different pytorch, cuda and python versions.

Triton uses `nvrtc` to compile kernels at runtime theoretically to comparable performance as CUDA. This converts this project into a pure python project which can be installed directly from a source distribution instead of platform and dependency-specific wheels.

The Triton reimplementation here is comparable to the CUDA one, it actually beats CUDA in the scenarios we most care about (low batches). See benchmarks in: https://github.com/torchmd/torchmd-net/issues/372#issuecomment-3660431497

What are your opinions @RaulPPelaez ? By the way many many thanks for the wonderful and extensive test suite! Without it this reimplementation would have been near impossible

Old CUDA benchmark performance
```
Benchmarking neighbor list generation for 1 batches with 64 neighbors on average
Summary
-------
N Particles Distance(ms)         Brute(ms)            Cell(ms)             
---------------------------------------------------------------------------
256         0.55  x1.00          0.01  x62.01         0.14  x3.85          
512         0.58  x1.00          0.01  x53.17         0.19  x3.11          
1024        0.65  x1.00          0.02  x35.45         0.19  x3.35          
2048        0.78  x1.00          0.04  x20.07         0.21  x3.64          
4096        1.05  x1.00          0.09  x11.99         0.26  x3.95          
8192        1.60  x1.00          0.22  x7.44          0.31  x5.14          
16384       2.76  x1.00          0.80  x3.46          0.52  x5.34          
32768       5.26  x1.00          6.54  x0.80          1.03  x5.09          
65536       11.37 x1.00          13.86 x0.82          1.98  x5.73          
131072      32.47 x1.00          35.11 x0.92          3.79  x8.58          
```

New Triton implementation performance
```
Benchmarking neighbor list generation for 1 batches with 64 neighbors on average
Summary
-------
N Particles Distance(ms)         Brute(ms)            Cell(ms)             
---------------------------------------------------------------------------
256         0.56  x1.00          0.01  x89.26         0.12  x4.67          
512         0.60  x1.00          0.01  x77.48         0.15  x3.86          
1024        0.64  x1.00          0.01  x50.94         0.13  x5.09          
2048        0.77  x1.00          0.03  x26.22         0.13  x5.93          
4096        1.04  x1.00          0.09  x11.51         0.14  x7.19          
8192        1.59  x1.00          0.33  x4.88          0.24  x6.57          
16384       2.75  x1.00          1.25  x2.20          0.32  x8.48          
32768       5.26  x1.00          4.86  x1.08          0.59  x8.94          
65536       11.37 x1.00          18.98 x0.60          1.10  x10.30         
131072      32.53 x1.00          75.24 x0.43          2.04  x15.97  
```

### DIFF
--- a/torchmdnet/extensions/triton_brute.py
+++ b/torchmdnet/extensions/triton_brute.py
@@ -186,7 +186,7 @@ def triton_neighbor_bruteforce(
         # Use stride 0 to broadcast single box to all batches (avoids CPU sync)
         box_batch_stride = 0 if box_vectors.size(0) == 1 else 9
 
-    neighbors = torch.full((2, max_num_pairs), -1, device=device, dtype=torch.int32)
+    neighbors = torch.full((2, max_num_pairs), -1, device=device, dtype=torch.long)
     deltas = torch.zeros((max_num_pairs, 3), device=device, dtype=dtype)
     distances = torch.zeros((max_num_pairs,), device=device, dtype=dtype)
     num_pairs = torch.zeros((1,), device=device, dtype=torch.int32)
--- a/torchmdnet/extensions/triton_cell.py
+++ b/torchmdnet/extensions/triton_cell.py
@@ -399,7 +399,7 @@ def triton_neighbor_cell(
     )
 
     # Allocate outputs
-    neighbors = torch.full((2, max_num_pairs), -1, device=device, dtype=torch.int32)
+    neighbors = torch.full((2, max_num_pairs), -1, device=device, dtype=torch.long)
     deltas = torch.zeros((max_num_pairs, 3), device=device, dtype=dtype)
     distances = torch.zeros((max_num_pairs,), device=device, dtype=dtype)
     num_pairs = torch.zeros((1,), device=device, dtype=torch.int32)


torchmdnet/models/output_modules.py
```diff
@@ -45,8 +44,11 @@ def pre_reduce(self, x, v, z, pos, batch):
         return
 
     def reduce(self, x, batch):
-        if not self.setup_for_compile:
-            is_capturing = x.is_cuda and is_current_stream_capturing()
+        if torch.jit.is_scripting():
+            # TorchScript doesn't support torch.cuda.is_current_stream_capturing()
+            self.dim_size = int(batch.max().item() + 1)
+        elif not self.setup_for_compile:
+            is_capturing = x.is_cuda and torch.cuda.is_current_stream_capturing()
```
RaulPPelaez: If I remember correctly, I introduced this because otherwise it was not possible to torch compile a model that was saved as torchscript (so for instance, one running via OpenMM-Torch). Are you checking for that?

stefdoerr:
Yes, nvrtc comes with pytorch/triton (through the nvidia-xxx pypi packages it pulls) so there is no need for installing nvrtc manually. You do need to pass the index URL though of pytorch which specifies the CUDA version you want to install but that was already the case before with pypi.

I am a bit disappointed that the implementation performs so bad with large batches because otherwise it would be a total win but I don't really see the use of such large batches for now.
Also I dropped the "shared" implementation because Triton doesn't provide low level memory access.

torchmdnet/models/output_modules.py
stefdoerr: Hm no, this is not something covered in the test suite? I have not run any further tests other than the ones in tests/ and the benchmarks.

RaulPPelaez:
I think you do fine by prioritizing the typical workloads you have now. You can always dig the old ones up shall the need arise. I also did not find a good way to do large number of batches with the cell list lol.

pyproject.toml
```diff
@@ -1,15 +1,27 @@
...
+requires-python = ">=3.9"
+dynamic = ["version"]
 classifiers = [
   "Programming Language :: Python :: 3",
   "Operating System :: POSIX :: Linux",
 ]
+dependencies = [
+  "h5py",
+  # "nnpops",
+  "torch==2.7.1",
```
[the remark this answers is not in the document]
stefdoerr: yes, of course, that was the idea I just forgot :) Done now

torchmdnet/models/output_modules.py
RaulPPelaez: I was actually referring to CUDA graphing. Sorry -.-
This is what the "is_cuda_graph_capturing" does and what is used in OpenMM-Torch. So a test like this but TorchScripting before graphing:
https://github.com/torchmd/torchmd-net/blob/75b16e625039a5e5e79c5e9d679c6c92ee274445/tests/test_model.py#L117-L161

BTW surprised only tensornet is graphable these days!

sef43:
I have checked using it in a full tensornet model (aceff1.0) - It looks to be correct, and speed is as reported, i.e. fast enough to not be the bottleneck when compared to the rest of the tensornet model. 

Before we merge I have two points that I want to check some more/should be discussed:

1.  when using it in a torch-scripted model (i.e in openmm-torch) it is only using the pure python code path, not the triton code path. (for small molecules this seems fast enough).  I think the triton will error out as it cannot be torch-scripted. 

3. I think the triton code will be recompiled each time the max_num_pairs (number of atoms) changes, this could significantly reduce speed when used as a single point calculator when looping over different molecules, and maybe also training with batch sizes with different sized molecules.

stefdoerr:
Very good point about the recompilation! I had not taken that into consideration. The kernels get recompiled whenever a `tl.constexpr` changes which indeed for brute was also number of atoms and cutoffs. I removed now the `tl.constexpr` from all those variables and kept it just for `use_periodic, loop, include_transpose` which means we will get max 8 kernels for all the combinations of those parameters.

Here the performance of the brute kernel after removing the constants related to number of atoms and cutoffs. Seems like it's negligible which is great!

Before the change (infinite kernels)
```
Benchmarking neighbor list generation for 1 batches with 64 neighbors on average
Summary
-------
N Particles Distance(ms)         Brute(ms)            Cell(ms)             
---------------------------------------------------------------------------
256         0.55  x1.00          0.01  x76.01         0.12  x4.52          
512         0.58  x1.00          0.01  x67.55         0.16  x3.73          
1024        0.65  x1.00          0.01  x48.34         0.13  x5.13          
2048        0.78  x1.00          0.03  x25.79         0.13  x5.97          
4096        1.04  x1.00          0.09  x11.48         0.14  x7.20          
8192        1.60  x1.00          0.33  x4.91          0.24  x6.58          
16384       2.75  x1.00          1.33  x2.06          0.33  x8.47          
32768       5.26  x1.00          4.99  x1.05          0.56  x9.44          
65536       11.36 x1.00          18.94 x0.60          1.11  x10.27         
131072      32.46 x1.00          75.16 x0.43          2.07  x15.71   
```

After the change (8 kernels only)
```
Benchmarking neighbor list generation for 1 batches with 64 neighbors on average
Summary
-------
N Particles Distance(ms)         Brute(ms)            Cell(ms)             
---------------------------------------------------------------------------
256         0.56  x1.00          0.01  x70.62         0.12  x4.65          
512         0.59  x1.00          0.01  x63.71         0.15  x3.89          
1024        0.66  x1.00          0.01  x45.11         0.13  x5.27          
2048        0.78  x1.00          0.03  x25.06         0.13  x6.09          
4096        1.05  x1.00          0.09  x11.11         0.15  x7.19          
8192        1.60  x1.00          0.33  x4.83          0.24  x6.58          
16384       2.76  x1.00          1.31  x2.11          0.34  x8.02          
32768       5.27  x1.00          4.94  x1.07          0.61  x8.65          
65536       11.38 x1.00          19.18 x0.59          1.22  x9.36          
131072      32.55 x1.00          75.71 x0.43          2.32  x14.06         
```

torchmdnet/extensions/triton_brute.py
```diff
@@ -0,0 +1,260 @@
...
+        num_pairs = torch.zeros((1,), device=device, dtype=torch.int32)
+
+        # Compute triangular pair count: n*(n-1)/2 without self-loops, n*(n+1)/2 with
+        if loop:
+            num_all_pairs = n_atoms * (n_atoms + 1) // 2
+        else:
+            num_all_pairs = n_atoms * (n_atoms - 1) // 2
+
+        # Grid covers only triangular pairs (not n²)
+        grid = lambda meta: (triton.cdiv(num_all_pairs, meta["BLOCK"]),)
```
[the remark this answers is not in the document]
stefdoerr: Great point! Fixed now

stefdoerr:
Okay I'll try working a bit more on it locally to see if I can get it to torch.export correctly with cell lists / triton code without spamming the PR. I'll let you know when I have an update.

sef43:
Training speed also seem to be the same which is good! 

(if I checkout the code before https://github.com/torchmd/torchmd-net/pull/373/commits/9179ad72a6c4a0af17d685cf2c9f8abe2d9dc7a9 then training speed was impacted by a factor of 2 due to the repeated kernel recompilations whenever the number of atoms changed)

stefdoerr:
Awesome, thanks for the benchmarking @sef43. I would also vote to merge without the export, mostly because the PR is growing too much to be reviewable. I gave it another look now and it looks good to go from my side. @RaulPPelaez @peastman feel free to give it a last look / comments

sef43 (approved):

PR #375, opened 2025-12-30
Add Tensornet2 architecture

sef43:
- adds the new tensornet2 architecture for the aceff-2.0 pre-trained model.

- the key addition is Aimnet2 style neutral charge equilibration and a coulomb interaction.

- This PR adds the tensonet2 model and a new output model that does a coulomb interaction using the predicted charges.

Note that the parent branch is #373

### DIFF
--- a/examples/aceff_examples/aceff.py
+++ b/examples/aceff_examples/aceff.py
@@ -8,7 +8,6 @@ from huggingface_hub import hf_hub_download
 
 model_file_path = hf_hub_download(
     repo_id="Acellera/AceFF-2.0",
-    revision="refs/pr/1",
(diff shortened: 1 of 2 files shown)


sef43:
It also includes some Tensornet optimizations first attempted in this PR: https://github.com/torchmd/torchmd-net/pull/320
which reduce memory by ~30% and increase speed by ~30%, where possible:
- I is kept as a scalar with shape [...,F]
- A is kept as a vector with shape [...,F,3]
- S is still a tensor with shape [...,F,3,3]

where F is the embedding dimension and the leading dim is the number of nodes or number of edges.

PR #390, opened 2026-03-31
workaround for compile bug

sef43:
this is a workaround for the issue reported here: https://github.com/openmm/openmm-ml/issues/136
which is a torch compile bug: https://github.com/pytorch/pytorch/issues/178871

Essentially when the model gets torch compiled if the number of atoms is not a multiple of 2 it can end up with an offset (i think deterministic) to the total energy from buggy generated triton code. The forces are not affected, relative energies after compilation are also not affected. The error appears when comparing energy from a non-compiled model with the energy from a compiled model.


sef43:
The failures look to be from torchscript

